PR #3, opened 2021-06-08
Add basic demonstration program for using CrateDB with vanilla Npgsql

amotl:
Hi there,

in order to improve the documentation at [1] and support the discussion at [2], this patch adds a basic example program based on vanilla Npgsql. It has been derived from [3].

With kind regards,
Andreas.

[1] https://crate.io/docs/npgsql/en/latest/connect.html
[2] https://community.crate.io/t/how-to-connect-npgsqldatabaseinfo-for-crate-database-in-c/698
[3] https://github.com/crate/crate-qa/tree/9e13566/tests/client_tests/stock_npgsql


mfussenegger:
Would be cool if we could connect this somehow with the efforts in crate-qa. To ensure the examples are tested and keep working with newer versions.

amotl:
> Would be cool if we could connect this somehow with the efforts in crate-qa. To ensure the examples are tested and keep working with newer versions.

I was thinking about the same but wasn't sure you would like to have the slightly enhanced version (at least from the perspective of an average user) replace the original variant in `crate-qa`. But as I see you are signaling interest in fusing those efforts, I will think about it how to implement this appropriately.

Currently, the variant in `cratedb-examples` has a bunch of inline comments but lacks most of the assert statements of the original variant in `crate-qa`. Instead, it has a more verbose output on stdout to give the average user some kind of visual feedback.

I believe we can figure out a way to address both needs within a single program in order not to duplicate the efforts in this area.

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
...
+    docker run -it --rm --publish=4200:4200 --publish=5432:5432 crate:4.5.1
+
+For building and running the demonstration program and connect to
+``localhost:5432``, invoke::
+
+    dotnet run 
+
+In order to connect to CrateDB Cloud, use::
+
+    dotnet run -- --host=clustername.aks1.westeurope.azure.cratedb.net --ssl-mode=Require --username=foobar --password='X8F%Shn)TESTvF5ac7%eW4NM'
```
hammerhead, line 33: Probably doesn't do any harm, but how about replacing the password with a placeholder as well?

spikes/npgsql-vanilla/Makefile
```diff
@@ -0,0 +1,12 @@
+help:
+	@echo "Please choose one of the following targets: start-cratedb, build, run"
+	@exit 2
+        
+start-cratedb:
+	docker run -it --rm --publish=4200:4200 --publish=5432:5432 crate:4.5.1 -C=cluster.routing.allocation.disk.threshold_enabled=false
```
hammerhead, line 6: Just out of curiosity, is there a particular reason the `cluster.routing.allocation.disk.threshold_enabled` parameter gets set here but isn't present in the `docker run` example given in README.rst?

amotl: That's because I regularly run into issues on my machine outlined at https://github.com/crate/crate/issues/11458.

spikes/npgsql-vanilla/README.rst
amotl: I thought it would be a good idea to also demonstrate how a regular CrateDB Cloud password looks like. If you think this is not appropriate, I can adjust this.

spikes/npgsql-vanilla/DemoProgram.cs
```diff
@@ -0,0 +1,176 @@
...
+            using (var cmd = new NpgsqlCommand("INSERT INTO testdrive.basic (x) VALUES (@x)", conn))
+            {
+                cmd.Parameters.AddWithValue("x", -999);
+                cmd.ExecuteNonQuery();
+            }
+
+            // Insert multiple data points.
+            await using (var cmd = new NpgsqlCommand("INSERT INTO testdrive.basic (x) VALUES (@x)", conn))
+            {
+                await using (var transaction = conn.BeginTransaction())
```
proddata, line 55: Should we even include transactions (basically not supported in CrateDB)

proddata:
Also would it not make more sense do include it in the docs, as people know often get drawn to by the npgsql fork? 

@mfussenegger Is time already to remove it?

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
+#################################################################
```
norosa, line 1: we should probably use the standard syntax for titles and headers (here and throughout the rest of the file)

see https://github.com/crate/crate-docs/blob/main/style/rst.rst#titles-and-headings

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
...
+
+The file ``DemoProgram.cs`` is a demonstration program written in C# which
+outlines typical, very basic usage scenarios with CrateDB.
+ 
+
+*****
+Usage
+*****
+
+In order to invoke a CrateDB instance for evaluation purposes, run::
```
norosa, line 22:
```suggestion
To invoke a CrateDB instance for evaluation purposes, run::
```

you can simplify here

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
...
+
+    docker run -it --rm --publish=4200:4200 --publish=5432:5432 crate:4.5.1
```
norosa, line 24: our standard syntax for console examples is:

```
.. code-block:: console

    sh$ echo foo
```

(here and elsewhere in the file)

spikes/npgsql-vanilla/Makefile
proddata: I still would keep it out :/

spikes/npgsql-vanilla/README.rst
norosa, line 33: this might display better if the command is wrapped over multiple lines with `\` characters

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
...
+For building and running the demonstration program and connect to
+``localhost:5432``, invoke::
+
+    dotnet run 
+
+In order to connect to CrateDB Cloud, use::
+
+    dotnet run -- --host=clustername.aks1.westeurope.azure.cratedb.net --ssl-mode=Require --username=foobar --password='X8F%Shn)TESTvF5ac7%eW4NM'
+
+For exploring all available options, run::
```
norosa, line 35:
```suggestion
For all available options, run:: 
```

you can simplify here

spikes/npgsql-vanilla/DemoProgram.cs
amotl: I believe this is here to demonstrate that this will act as a no-op instead of croaking.

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
...
+In order to invoke a CrateDB instance for evaluation purposes, run::
+
+    docker run -it --rm --publish=4200:4200 --publish=5432:5432 crate:4.5.1
+
+For building and running the demonstration program and connect to
+``localhost:5432``, invoke::
+
+    dotnet run 
+
+In order to connect to CrateDB Cloud, use::
```
norosa, line 31:
```suggestion
To connect to CrateDB Cloud, use::
```

(as above)

spikes/npgsql-vanilla/README.rst
```diff
@@ -0,0 +1,40 @@
+#################################################################
+Basic demonstration program for using CrateDB with vanilla Npgsql
+#################################################################
+
+
+************
```
norosa, line 6: because this is a short document with only one substantial section ("Usage") it is probably best to drop this heading

we try to include a few intro paras right after the title in most of our RST documents. "Introduction" sections are then only needed if there is a need to go into more detail

spikes/npgsql-vanilla/README.rst
amotl: Thank you!

a) With 5960fa3f, I added `.. highlight: console`, which will essentially do the same job, right?
b) Do you really think we should keep adding prompts like `sh$ ` everywhere?

spikes/npgsql-vanilla/README.rst
amotl: Thank you. Addressed with 5960fa3.

spikes/npgsql-vanilla/README.rst
amotl: Thank you. Addressed with 5960fa3.

amotl:
> Also would it not make more sense do include it in the docs, as people know often get drawn to by the npgsql fork?

Right and thanks! The work here was only the foundation for https://github.com/crate/crate-npgsql/pull/58.

spikes/npgsql-vanilla/DemoProgram.cs
proddata: Ah ... now I get it ... I am wondering if we really want to use the full test for a simple example.
Considering also that the whole connection part is on the bottom :/

spikes/npgsql-vanilla/README.rst
amotl: Thank you. ca0e6edc addresses this.

spikes/npgsql-vanilla/README.rst
amotl: See ca0e6edc.

spikes/npgsql-vanilla/README.rst
amotl: Addressed with ca0e6edc. Thanks!

spikes/npgsql-vanilla/DemoProgram.cs
mfussenegger: Afaik using the transaction is important to get efficient bulk insert execution. But would need to double check the message flow.

mfussenegger:
> Also would it not make more sense do include it in the docs, as people know often get drawn to by the npgsql fork?
> 
> @mfussenegger Is time already to remove it?

From my perspective stock npgsql is better supported. At least there we have some basic tests as part of the nightly crate-qa runs. That's not the case for the fork + plugin.

spikes/npgsql-vanilla/README.rst
norosa: a) yes, great! however, the highlight directive should appear on the first line of the file (for consistency with the rest of our docs)

b) the `sh$` prompt was designed for consistency with the frequent use of `cr>` prompts in the main docs. we can revisit that decision if you like, but I think that should be done separately (perhaps with a PR to crate-docs/style)

spikes/npgsql-vanilla/Makefile
amotl: All right. Addressed with f0cb179c912ce.

spikes/npgsql-vanilla/README.rst
amotl: All right. Addressed with f0cb179c912ce and https://github.com/crate/crate-docs/issues/86.

amotl:
@mfussenegger: Following up on my arguments at https://github.com/crate/cratedb-examples/pull/3#issuecomment-856831710, I tried to address both aspects within the same machinery with f97912a7aa2:

> I believe we can figure out a way to address both needs within a single program in order not to duplicate the efforts in this area.

The demo program can now both be used interactively and also invoked as an xUnit test suite. That second aspect should satisfy its usability as a qa program.

amotl:
Dear @mfussenegger,

this patch would be ready from my perspective. Please let me know if we should submit it to https://github.com/crate/crate-qa/tree/master/tests/client_tests/stock_npgsql instead.

With kind regards,
Andreas.

.github/workflows/tests-npgsql.yml
```diff
@@ -0,0 +1,73 @@
...
+      matrix:
+        os: [ubuntu-latest]  # , macos-latest
+        dotnet-version: ['3.1.x', '5.0.x']  # 6.0.0-preview.4
+        cratedb-version: ['4.3.2', '4.4.3', '4.5.1', 'nightly']
+        npgsql-version: ['5.0.5', '6.0.0-preview4']
+        #include:
+        #- dotnet-version: 3.1.x
+        #  dotnet-framework: netcoreapp3.1
+        #- dotnet-version: 5.0.x
+        #  dotnet-framework: net5.0
```
mfussenegger, line 33:
```suggestion
```

.github/workflows/tests-npgsql.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Set up .NET ${{ matrix.dotnet-version }}
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: ${{ matrix.dotnet-version }}
+
+      - name: Invoke tests on netcoreapp3.1
+        if: matrix.dotnet-version == '3.1.x'
+        run: |
+          cd spikes/npgsql-vanilla
+          dotnet test --framework=netcoreapp3.1
```
mfussenegger, line 55: Wouldn't something like `dotnet test --framework=${{ matrix.dotnet-version }}` work to avoid the conditionals?

spikes/npgsql-vanilla/Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+build:
+	dotnet build
+
+run:
+	dotnet run --framework=${DOTNET_FRAMEWORK}
+
+test:
+	dotnet test --framework=${DOTNET_FRAMEWORK} --verbosity=minimal
+	@#dotnet test --verbosity=normal
```
mfussenegger, line 18:
```suggestion
```

mfussenegger: I'd  remove the whole Makefile, no need to build a custom make based dsl around the dotnet command. Imho it obfuscates what is happening and doesn't offer any advantage. Dotnet developers will know how to use the `dotnet` cli, they won't know what our custom make targets are and will have to learn something extra if they look at this.

But I'll leave the decision to you.

.github/workflows/tests-npgsql.yml
amotl: Unfortunately not, because we would need a map to associate the .NET framework identifier used by GHA's `actions/setup-dotnet` (3.1.x, 5.0.x) to the identifier used by the `dotnet` command's argument `--framework` (netcoreapp3.1, net5.0).

I tried to do it with the `include:` attribute on GHA's matrix definition but failed on that.

.github/workflows/tests-npgsql.yml
amotl: Thanks for spotting this remnant. It was a leftover from my attempt to make the configuration smarter, see next comment. I've removed it on behalf of fc64076aa.

spikes/npgsql-vanilla/Makefile
amotl: Thank you. I moved remaining "notes" from the Makefile into the corresponding README.rst instead and removed the Makefile afterwards, see fc64076aa.

mfussenegger (approved):

PR #8, opened 2022-12-06
Add basic example with Ruby

amotl:
This patch supports https://github.com/crate/crate_ruby/pull/39.


spikes/ruby-basic/Gemfile
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
+
+# Specify your gem's dependencies in crate_ruby.gemspec
+gemspec
```
hammerhead, line 6: I think a gemspec is not needed here and you can put dependencies directly into the Gemfile. A gemspec is only used when packaging and distributing a gem, but not for standalone scripts/applications.

hammerhead (approved):

PR #9, opened 2023-01-20
Java/jOOQ: Add basic demo application and testing rig

amotl:
Hi there,

after learning that evaluating [CrateDB](https://github.com/crate/crate) together with [jOOQ](https://github.com/jOOQ/jOOQ) is a long cherished wish, this patch finally stages a corresponding testing rig.

The program and test suite can be invoked like:
```
./gradlew run
./gradlew test
```

With kind regards,
Andreas.

/cc @hlcianfagna, @matriv, @mkleen, @proddata, @seut, @hammerhead 

### About
The idea of jOOQ is to generate typesafe code based on the SQL schema. Then, accessing a database table using the [jOOQ DSL API](https://www.jooq.org/doc/latest/manual/sql-building/dsl-api/) looks like this:

https://github.com/crate/cratedb-examples/blob/278b0f8ad123052c690ccebdce2a064cbc49ce9b/by-language/java-jooq/src/main/java/io/crate/demo/jooq/Application.java#L110-L115

This feature currently does not work with CrateDB yet. The code provided within the `src/generated` directory has _not_ been derived by reflecting the database schema from CrateDB. In order to make it work out of the box using jOOQ's PostgreSQL dialect, CrateDB would need to support `WITH RECURSIVE` CTEs.

- https://github.com/crate/crate/issues/12544
- GH-10


mkleen:
@amotl Are you sure that you want to check in autogenerated code ? Would it be easier to generate the code add-hoc as part of the build process ?

amotl:
> Are you sure that you want to check in autogenerated code ? Would it be easier to generate the code add-hoc as part of the build process?

@mkleen: It is currently not possible to generate the corresponding code, see the "Caveat" section above. However, I will do it quickly if you can fix https://github.com/crate/crate/issues/12544 right away?

mkleen:
> > Are you sure that you want to check in autogenerated code ? Would it be easier to generate the code add-hoc as part of the build process?
> 
> @mkleen: Thanks for your suggestion. However, it is currently not possible to generate the corresponding code, see the "Caveat" section above. However, I will do it quickly if you can fix [crate/crate#12544](https://github.com/crate/crate/issues/12544) right away? ;]

I am referring to the generated domain model such as 
https://github.com/crate/cratedb-examples/pull/9/files#diff-8daa01076532b457e57ee3b7ae90d24cf8ff5f1505e155d9777cc151299181efR18

mkleen:
> > > Are you sure that you want to check in autogenerated code ? Would it be easier to generate the code add-hoc as part of the build process?
> > 
> > 
> > @mkleen: Thanks for your suggestion. However, it is currently not possible to generate the corresponding code, see the "Caveat" section above. However, I will do it quickly if you can fix [crate/crate#12544](https://github.com/crate/crate/issues/12544) right away? ;]
> 
> I am referring to the generated domain model such as
> 
> https://github.com/crate/cratedb-examples/blob/eb951b9635a56a5c827516159f1c44051459fd52/by-language/java-jooq/src/generated/java/io/crate/demo/jooq/model/tables/records/AuthorRecord.java#L18-L23

amotl:
> It is currently not possible to generate the corresponding code.

jOOQ runs a CTE with `WITH RECURSIVE` for schema reflection. We are not there yet.

mkleen:
> > It is currently not possible to generate the corresponding code.
> 
> jOOQ runs a CTE with `WITH RECURSIVE` for schema reflection. We are not there yet.

Ok, now i get it. The sources are not generated from CrateDB. I am not sure if this a good idea to provide this as a sample app then if it's not possible to use straight away.

mkleen:
What's the best practice with jOOQ ? Is it possible to use it without source generation with a normal domain model ?

amotl:
> I am not sure if this a good idea to provide this as a sample app then if it's not possible to use straight away.

Why so serious? Here we go, the intention is clearly stated in the README:

> It is intended as a basic example to demonstrate what currently works, and as a testing rig for eventually growing a full-fledged CrateDB dialect, or at least making the code generator work. Contributions are welcome.

amotl:
> What's the best practice with jOOQ? Is it possible to use it without source generation with a normal domain model?

This patch is here to demonstrate just that, right. And to open the door for subsequent refinements.

mkleen:
> > What's the best practice with jOOQ? Is it possible to use it without source generation with a normal domain model?
> 
> This patch is here to demonstrate just that, right. And to open the door for subsequent refinements. Best practice, and main intention, is definitively to use it with code generation, according to upstream docs. Personally, I think the other details of jOOQ are equally powerful, even without using code generation.

I think if we provide an example with CrateDB and jOOQ then this should work out of the box. If the standard workflow is to generate the domain model from the database schema then this should work too. Imagine to explain to a user that he first has to generate the domain model from another database and then modify it by hand to make it work with CrateDB.

amotl:
> I am so serious about that, because I am tired of example apps which don't follow best practices. They are painful to maintain and also painful for the users because they will start on a weak foundation.

I am not sure that committing the generated domain model is not following best practices in this particular case. In order to generate the code, you will need a _running_ database upfront. That would defeat using the domain model within unit tests which use mocking-based techniques right away, essentially rendering the whole application test suite completely useless.

When assembling this code, I followed the same practice of providing the generated domain model classes within the source code repository like seen at [^1]. This was invaluable information for me to get started directly using CrateDB, without being able to use the code generator.

I think it is a good start into this topic and will help others to get started quickly as well. I think it does not mislead the user in any way, and follows the style of the example code as provided by the upstream author(s). I would clearly like to see any gaps being closed on behalf of subsequent iterations, in order to leverage more of what jOOQ has to offer.

[^1]: https://github.com/jOOQ/jOOQ-mcve/tree/main/jOOQ-mcve-java/src/main/java/org/jooq/mcve/java

amotl:
> Imagine to explain to a user that he first has to generate the domain model from another database and then modify it by hand to make it work with CrateDB.

Please also have a look at the "dynamic schema" example code, which also demonstrates a well-supported subset of jOOQ's feature set.

https://github.com/crate/cratedb-examples/blob/eb951b9635a56a5c827516159f1c44051459fd52/by-language/java-jooq/src/main/java/io/crate/demo/jooq/Application.java#L132-L191

mkleen:
Ok, seems like the sample app from jooq also checks in the generated sources 
https://github.com/jOOQ/jOOQ-mcve/tree/main/jOOQ-mcve-java/src/main/java/org/jooq/mcve/java

proddata:
> However, I will do it quickly if you can fix https://github.com/crate/crate/issues/12544 right away? ;]

Out of curiosity, is this the only missing feature?
Also is this coming from the pg jdbc driver? have you tried the crate variant?

amotl:
> Also wondering if generation would work, if specified as `org.jooq.meta.jdbc.JDBCDatabase`?

Thanks for the suggestion, I will try.

amotl:
>> Also wondering if code generation would work when using the `org.jooq.meta.jdbc.JDBCDatabase` generator?
>
> Thanks for the suggestion, I will try.

It also croaks, see https://github.com/crate/cratedb-examples/pull/10#issuecomment-1398655879.

mkleen:
@amotl I spent a bit more time on reading on jOQL. There is the option to use it without code generation. I think I would prefer if we set up an example for jOQL to use an option which the user can reproduce without any hacks. I would also spend a little bit more time on advanced functionality to see if this works end-to-end. Let's make sure this ages well.

mkleen:
However, if https://github.com/crate/crate/issues/12544 is the blocker to jOQl support fully working, I would probably add an issue and then wait for feedback from the community. Then we can support the feature once it gets upvotes.

amotl:
Hi @mkleen,

thanks for your feedback. Maybe GH-10 already addresses some of your suggestions. I hope that this patch will only be the baseline, and would like to evaluate other features and more advanced operations than just the basic ones demonstrated here, on behalf of subsequent patches.

With kind regards,
Andreas.

by-language/java-jooq/src/generated/java/io/crate/demo/jooq/model/DemoDatabase.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * This class is generated by jOOQ.
+ */
+@Generated(
+        value = {
+                "https://www.jooq.org",
+                "jOOQ version:3.17.7"
+        },
+        comments = "This class is generated by jOOQ."
+)
```
amotl, line 28: @seut mentioned that the pseudo-generated classes should not contain comments like this.

Instead, the comments should be more clear about that they have been hand-written / derived from other examples, eventually citing the README:

https://github.com/crate/cratedb-examples/blob/eb951b9635a56a5c827516159f1c44051459fd52/by-language/java-jooq/README.rst?plain=1#L46-L52

mkleen: How would a solution look like without generated code ? Would that result in the same kind of classes ?

seut: @mkleen Afaik this can be seen here https://github.com/crate/cratedb-examples/blob/eb951b9635a56a5c827516159f1c44051459fd52/by-language/java-jooq/src/main/java/io/crate/demo/jooq/Application.java#L132.

amotl: GH-11 takes a better approach on this. Instead of leaving us or any reader hanging on this detail, it uses the existing SQL DDL scripts to reflect the database schema, see _DDLDatabase: Code generation from SQL files_ [^1]. Let me know what you think about it.

As outlined within the upstream documentation, and accordingly cited within the README now, this fits well with the very reasonable scenario that your SQL DDL statements are usually maintained in form of multiple incremental migration scripts anyway, when writing an application where schema changes are important to be tracked within the source code repository. Within this scenario, there is no need to connect to an existing database for reflecting its schema.

jOOQ even promises it could cope well with CrateDB-specific extensions to the `CREATE TABLE` DDL statements, which definitively matters.

> While the script uses pretty standard SQL constructs, you may well use some vendor-specific extensions, and even [DML statements](https://www.jooq.org/doc/latest/manual/sql-building/sql-statements/) in between to set up your schema - it doesn't matter. By [ignoring unsupported content](https://www.jooq.org/doc/latest/manual/code-generation/codegen-ddl/#ignoring-unsupported-content), the [jOOQ SQL parser](https://www.jooq.org/doc/latest/manual/sql-building/sql-parser/) supports parsing everything that is representable through the jOOQ API, as well as some well known vendor-specific syntax, [and even unknown vendor-specific syntax by adding correspdonding hints].

[^1]: https://www.jooq.org/doc/latest/manual/code-generation/codegen-ddl/

PR #19, opened 2023-02-21
Naming things: Improve folder layout by renaming `spikes` to `stacks`

amotl:
As discussed recently, this brings a better naming to the folder layout, using `stacks` for the "more grown-up usage scenarios" folder, also being a kind of staging area for potential "reference architecture" types of documentation.

/cc @seut


matriv (commented):
Thank you!

PR #20, opened 2023-02-21
Kafka/Flink: Modernize sandbox environment

amotl:
### About
Modernize the test rig for [running "Apache Kafka, Apache Flink and CrateDB"](https://github.com/crate/cratedb-examples/blob/main/stacks/kafka-flink/README.rst).

### Details
- Use Docker Compose v2
- `kafkacat` has been renamed to `kcat`
- Manage software component versions in `.env` file
- Manage project tasks within Docker Compose file
- Improve documentation

### Outlook
A little CI setup, and actually bumping the software component versions, will happen by subsequent patches.

PR #21, opened 2023-02-23
Kafka/Flink: Add end-to-end test

amotl:
Hi again,

as promised at GH-20, this patch adds a basic end-to-end test for the Kafka/Flink/CrateDB stack, also running it on CI/GHA. I am sad that it is a Bash script, but this is still the most compact yet coherent representation of the recipe. The outcome will be a description like this, in the error case with an exitcode != 0.

```
=========================
Verifying data in CrateDB
=========================
SUCCESS: Database table contains expected number of 5000 records.
```

```
=========================
Verifying data in CrateDB
=========================
ERROR:   Expected database table to contain 5000 records, but it contains 0 records.
```

With kind regards,
Andreas.

### DIFF
--- a/.github/workflows/test-kafka-flink.yml
+++ b/.github/workflows/test-kafka-flink.yml
@@ -1,0 +1,42 @@
+name: Apache Kafka, Apache Flink, and CrateDB
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+    - '.github/workflows/test-kafka-flink.yml'
+    - 'stacks/kafka-flink/**'
+  pull_request:
+    branches: [ main ]
+    paths:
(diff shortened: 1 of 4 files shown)


matriv (approved):
wow! Thx, you really put effort into this!

amotl:
`verify-data` is currently a bit thin. It only verifies that the right amount of records has made it to CrateDB.

https://github.com/crate/cratedb-examples/blob/6f5df73d6f4204df395adc20e99df0ed79b0f642/stacks/kafka-flink/test.sh#L121-L135

For future tests, especially when improving the data mapping with upcoming changes, I'd recommend to add a small Python-based test file, which will exercise data verifications within CrateDB in more detail.

PR #23, opened 2023-02-23
Kafka/Flink: Update software versions

amotl:
Update the software components to those versions:

- Confluent Kafka 6.1.9
- CrateDB 5.2.2
- Flink 1.13

We currently can't go higher than Flink 1.13, because it will not invoke Flink jobs compiled with/for Flink 1.12, see [^1][^2].

[^1]: https://github.com/crate/cratedb-flink-jobs
[^2]: https://github.com/crate/cratedb-flink-jobs/blob/0.2/build.gradle#L24


stacks/kafka-flink/README.rst
```diff
@@ -87,9 +90,13 @@ test progam repeatedly. Use the ``--keepalive`` option for that::
 In order to run specific subcommands/functions defined within the file, invoke,
 for example::
 
-    bash test.sh teardown
     bash test.sh stop-services
 
+If, by chance, the ``verify-data`` test step croaked, resources are currently
```
matriv, line 95:
```suggestion
If, by chance, the ``verify-data`` test step fails, resources are currently
```

matriv (approved):
Once again, thx for the effort, left a comment

PR #24, opened 2023-03-23
Java/JDBC: Adjustments to also demonstrate usage with CrateDB Cloud

amotl:
Hi there,

this patch aims to improve the canonical Java/JDBC example to also cover demonstrating usage with CrateDB Cloud.

- Add [JCommander](https://jcommander.org/) for command line argument parsing
- Skip provisioning a database schema, query `sys.summits` instead
- Improve error handling

With kind regards,
Andreas.

### DIFF
--- a/by-language/java-jdbc/BasicPostgresCrateDB.java
+++ b/by-language/java-jdbc/BasicPostgresCrateDB.java
@@ -5,50 +5,70 @@ import java.sql.DriverManager;
 import java.sql.ResultSetMetaData;
 import java.util.Locale;
 import java.util.Properties;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.JCommander;
 
 public class BasicPostgresCrateDB {
 
+    @Parameter(names = {"--dburl", "-d"}, description = "Database URL")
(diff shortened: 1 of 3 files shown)


by-language/java-jdbc/BasicPostgresCrateDB.java
```diff
@@ -5,50 +5,70 @@
...
+        String connectionUrl = dburl;
 
         Properties connectionProps = new Properties();
-        // https://jdbc.postgresql.org/documentation/head/connect.html
-        connectionProps.put("user", "crate");
-        connectionProps.put("password", "");
-        connectionProps.put("ssl", false);
+        connectionProps.put("user", user);
+        connectionProps.put("password", password);
+        //connectionProps.put("ssl", "true");
```
amotl, line 53: Apparently, setting the `ssl` parameter is not needed. The driver itself seems to figure out how to correctly connect to both a non-SSL protected instance on localhost, and an SSL-protected instance on CrateDB Cloud, without needing to configure it.

hlcianfagna (approved):

PR #46, opened 2023-03-25
stacks/kafka-flink: Also use the rig with CrateDB Cloud easily

amotl:
This patch adjusts the infrastructure and adds corresponding notes how to invoke the test scenario on a CrateDB Cloud instance. The rendered version of the updated documentation section can be inspected at [README » Usage » CrateDB Cloud](https://github.com/crate/cratedb-examples/blob/1329eaf56/stacks/kafka-flink/README.rst#cratedb-cloud).


stacks/kafka-flink/test.sh
```diff
@@ -97,7 +96,7 @@ function feed-data() {
   echo
 
   # Wait a bit for the data to transfer and converge successfully.
-  sleep 1
+  sleep 2
```
matriv, line 99: you can make it 5 to be on the safe side.

matriv (approved):

stacks/kafka-flink/test.sh
amotl: I've increased to to three seconds. Unless we get errors, I think it is fine.

matriv (approved):

PR #84, opened 2023-09-06
MLflow: Add example program and guidelines

amotl:
## About

We have been looking at making [MLflow Tracking](https://mlflow.org/docs/latest/tracking.html) work together with CrateDB.

- https://github.com/crate/crate-clients-tools/issues/47

Our [fork of MLflow](https://github.com/crate-workbench/mlflow) now includes corresponding patches to roughly make it work. It still needs to be validated thoroughly.

This patch adds an example program, which effectively represents the ML experiment, and corresponding guidelines how to set up MLflow and run it. It is derived from excellent resources provided by @andnig - thank you so much!

## Preview

Documentation/Walkthrough, rendered by GitHub, can be inspected at [MLflow and CrateDB](https://github.com/crate/cratedb-examples/blob/amo/stack-mlflow/framework/mlflow/readme.md).

## Backlog

- [x] Demonstrate how to use that exercise with CrateDB Cloud.
- [ ] Demonstrate how to use that exercise with the Databricks backend.
- [x] Maybe conceive `mlflow-cratedb` using monkeypatching, in order to ship it as a little Python package, wrapping MLflow correspondingly.
- [ ] Add test coverage on behalf of a little pytest test case, and a corresponding GHA workflow.

/cc @andnig, @karynzv, @marijaselakovic, @hammerhead, @ckurze


stacks/mlflow/tracking_merlion.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    # Connecting to crate
+    conn = connect_database()
+    try:
+        conn.cursor().execute("SELECT * FROM machine_data LIMIT 1;")
+        return
+    except ProgrammingError as ex:
+        if "Relation 'machine_data' unknown" not in str(ex):
+            raise
```
seut, line 40: Minor: Looking at the related system classes to check table existence would be safer than expecting a concrete exception:
```suggestion
    cursor = conn.cursor()
    cursor.execute("SELECT table_name FROM information_schema.tables WHERE table_name = 'machine_data'  and table_schema = 'doc'");
    if cursor.rowcount > 0:
      raise Exception("Relation 'machine_data' already exists")
```

stacks/mlflow/readme.md
```diff
@@ -0,0 +1,142 @@
...
+
+### CrateDB on localhost
+
+In order to spin up a CrateDB instance without further ado, you can use
+Docker or Podman.
+```shell
+docker run --rm -it --publish=4200:4200 --publish=5432:5432 \
+  --env=CRATE_HEAP_SIZE=4g crate \
+  -Cdiscovery.type=single-node \
+  -Ccluster.routing.allocation.disk.threshold_enabled=false
```
seut, line 71: I don't think that its a good idea to advertise disabling this threshold in examples. If disabled, writing data into CrateDB can lead to a machine crash due out-of-disk-space issues.

amotl: You are right. That slipped in accidentally.

stacks/mlflow/tracking_merlion.py
amotl: Introspecting the information schema tables is definitively a better way to handle this. Thanks!

amotl: I revisited this detail, and came to a different conclusion, see https://github.com/crate-workbench/mlflow-cratedb/pull/13#discussion_r1320781162.

stacks/mlflow/readme.md
amotl: Fixed with a31c92856d, thanks again.

PR #86, opened 2023-09-20
Python: Add command-line program to work with the CrateDB BLOB store

amotl:
## About

The guideline at the Python client documentation about [working with the CrateDB BLOB store](https://cratedb.com/docs/python/en/latest/blobs.html) is sweet, but a runnable program is even better.

## Backlog

- [ ] Hasn't been verified with CrateDB Cloud yet.
- [ ] Wasn't used with larger amounts of data yet.
- [ ] In order to be used within own programs and libraries of CrateDB users,
       we may want to think about shipping it in one way or another,
       because it is a good building block.

/cc @andnig, @hlcianfagna, @hammerhead, @karynzv, @marijaselakovic, @proddata

### DIFF
--- a/by-language/python/blob.py
+++ b/by-language/python/blob.py
@@ -1,0 +1,255 @@
+"""
+## About
+Access CrateDB's BLOB store from the command-line.
+
+## Usage
+
+For convenient interactive use, define two environment variables.
+When not defining `--url` or `CRATEDB_HTTP_URL`, the program will
+connect to CrateDB at `crate@localhost:4200` by default.
+
+Synopsis::
+
+    # Define the HTTP URL to your CrateDB instance.
+    export CRATEDB_HTTP_URL=https://username:password@cratedb.example.net:4200/
+
+    # Define the BLOB container name.
+    export CRATEDB_BLOB_CONTAINER=testdrive
+
+    # Upload an item to the BLOB store.
+    python blob.py upload /path/to/file
+    418a0143404fb2da8a1464ab721f6d5fb50c3b96
+
+    # Download an item from the BLOB store.
+    python blob.py download 418a0143404fb2da8a1464ab721f6d5fb50c3b96
+
+Full command line example, without defining environment variables::
+
+    python blob.py \
+        --url=http://crate@localhost:4200/ --container=testdrive \
+        upload /path/to/file
+
+## References
+
+- https://cratedb.com/docs/python/en/latest/blobs.html
+- https://cratedb.com/docs/crate/reference/en/latest/general/blobs.html
+- https://cratedb.com/docs/crate/reference/en/latest/sql/statements/create-blob-table.html
+"""
+import io
+import logging
+import os
+import sys
+import typing as t
+from argparse import ArgumentError, ArgumentParser
+from pathlib import Path
+
+from crate import client
+from crate.client.blob import BlobContainer
+from crate.client.connection import Connection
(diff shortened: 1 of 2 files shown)


by-language/python/blob.py
```diff
@@ -0,0 +1,255 @@
...
+    parser = ArgumentParser()
+    url = parser.add_argument("-u", "--url", type=str)
+    container = parser.add_argument("-c", "--container", type=str)
+
+    actions = parser.add_subparsers(
+        dest="action",
+        title="action",
+        description="valid subcommands",
+        help="additional help",
+    )
```
amotl, line 184: Nit: Define `title`, `description`, and `help` better.

by-language/python/blob.py
```diff
@@ -0,0 +1,255 @@
...
+        return self
+
+    def __exit__(self, *excs):
+        self.disconnect()
+
+
+def setup_logging(level=logging.INFO):
+    """
+    What the function name says.
+    """
```
seut, line 135: Meaningless function doc?
```suggestion
```

by-language/python/blob.py
```diff
@@ -0,0 +1,255 @@
...
+    log_format = "%(asctime)-15s [%(name)-10s] %(levelname)-8s: %(message)s"
+    logging.basicConfig(format=log_format, stream=sys.stderr, level=level)
+
+
+def truncate(value: bytes) -> bytes:
+    """
+    Truncate long string.
```
seut, line 142:
```suggestion
    Truncate long string to be used for logging purpose.
```

by-language/python/blob.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def example(url: str, container_name: str):
+    """
+    An example conversation with the BLOB store (upload, download, delete).
+    """
+
+    # Define arbitrary content for testing purposes.
+    content = "An example payload.".encode("utf-8")
+    # content = b"\xde\xad\xbe\xef" * 256
```
seut, line 159: remove?

by-language/python/blob.py
seut: do it? ;)

by-language/python/blob.py
```diff
@@ -0,0 +1,255 @@
...
+            self.run_sql(f'CREATE BLOB TABLE "{self.name}";')
+        except ProgrammingError as ex:
+            if "RelationAlreadyExists" not in ex.message:
+                raise
+
+    def refresh(self):
+        """
+        Optionally synchronize data after write operations.
+        """
+        self.run_sql(f'REFRESH TABLE "{self.name}";')
```
seut, line 116: looks unused, remove it?
Otherwise I think it would have to use the `blob` schema, such as `REFRESH TABLE blob."{self.name}";`

amotl: Yeah, I also recognized this detail yesterday. Thank you. So, effectively, there is no way to run a BLOB table in a different schema? It is _always_ `blob`?

seut: Yes it's always `blob`, see also https://cratedb.com/docs/crate/reference/en/5.4/sql/statements/create-blob-table.html (syntax does not support to define a schema name).

amotl:
FYI: I've just reused that basic implementation over at https://github.com/crate-workbench/mlflow-cratedb/pull/33, adding more features to implement an object store.

PR #64, opened 2023-05-11
[Python/SQLAlchemy] Add example programs about efficient bulk inserts with SQLAlchemy, pandas, and Dask

amotl:
Hi there,

this patch carries over the example programs about Python/SQLAlchemy/pandas/Dask from https://github.com/crate/crate-python/pull/552. Originally, the corresponding programs have been conceived as individual gists:

- [SQLAlchemy insert gist]
- [pandas insert gist]
- [Dask insert gist]

With kind regards,
Andreas.

/cc @marijaselakovic, @karynzv, @WalBeh, @hammerhead, @hlcianfagna, @andnig, @ckurze 

[Dask insert gist]: https://gist.github.com/amotl/c8039291fb4b4b16879fd8814de9ecdf
[pandas insert gist]: https://gist.github.com/amotl/d39db8eff6ad5de5898ce61fac4c0af3
[SQLAlchemy insert gist]: https://gist.github.com/amotl/d0fa29ac52d4c10e4a2544cffd606dfc

### Backlog
- [x] ~~Software tests and Dependabot configuration works well for the other code example directories, and will be added here as well.~~ _Done, with a code coverage of 95%._
- [x] ~~https://github.com/crate/crate-python/pull/553 needs to be merged and released beforehand, to make the examples actually work.~~ _Merged in July 2023._
- [ ] Some adjustments may be added to bring all programs into the same shape.
- [ ] Other than demonstrating write operations, also demonstrate read operations?

=> Backlog items have been added to GH-95, in order to finally get this merged, so it can be promoted and re-used.

PR #96, opened 2023-10-28
ngr everywhere

amotl:
## About

Coming from https://github.com/crate/cratedb-examples/pull/64#pullrequestreview-1702806663, this patch intends to make `ngr` work on a few other projects as well.


by-language/php-pdo/.gitignore
```diff
@@ -0,0 +1 @@
+/vendor
```
matriv, line 1: Is this correct? shouldn't it be:
```suggestion
./vendor
```

.github/workflows/test-npgsql.yml
```diff
@@ -63,35 +65,15 @@ jobs:
...
-
-      - name: Display package information
-        run: |
-          dotnet restore
-          dotnet list . package
-
-      - name: Invoke tests on .NET 5
-        if: matrix.dotnet-version == '5.0.x'
-        run: |
-          dotnet test --framework=net5.0 --collect:"XPlat Code Coverage"
```
matriv, line 84: Apologies if I didn't understand correctly, do we loose now the tests with different platform versions?

matriv (commented):
Thx a lot @amotl I'm always rooting for consistency, left some comments.

.github/workflows/test-npgsql.yml
amotl: Nothing will be lost. All those stray commands have just been refactored into `testing/ngr.py`. It now has all the smartness "how to build for us" in one place.

by-language/php-pdo/.gitignore
amotl: I am not an expert, but I think the patterns are anchored from the point where it is located.

.github/workflows/test-npgsql.yml
```diff
@@ -29,22 +31,22 @@ concurrency:
 defaults:
   run:
     shell: bash
-    working-directory: by-language/csharp-npgsql
 
 jobs:
   test:
-    name: "CrateDB: ${{ matrix.cratedb-version }}
+    name: "
```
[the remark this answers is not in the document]
amotl: The name of the workflow is in the very first line `name: C# Npgsql`. Here, on the task level, we enumerate the test matrix slots.

by-language/php-pdo/.gitignore
matriv: It's minor really, just checked: https://git-scm.com/docs/gitignore
so:
```
For example, a pattern doc/frotz/ matches doc/frotz directory, but not a/doc/frotz directory; however frotz/ matches frotz and a/frotz that is a directory (all paths are relative from the .gitignore file).
```

So without anything, just `vendor` it's already relative to `.gitignore` dir.
I just questioned it, as I've never seen a leading `/` so far in a `.gitignore`.

.github/workflows/test-npgsql.yml
matriv: Yep, sorry, I missed: [dotnet_version_majorYep, sorry, I missed: dotnet_version_major](https://github.com/crate/cratedb-examples/pull/96/files#diff-75b364c2b98053dc2ce2711ebc3237f6e4e73ce920358bbd5e99fb37f8c58fcaR152)

matriv (approved):
Thx, but please wait for some other reviewer, more experienced at python.

by-language/php-pdo/.gitignore
amotl: Let's consider a typical PHP project as an example, which has this layout:

```
/vendor
/composer.lock
```

When I absolutely want to reflect that within a `.gitignore` file, I absolutely write that down there, no matter where the directory is located within a larger directory. It precisely means what is intending to describe. When, for example, just writing `vendor`, it will ignore all `vendor` folders within all other subdirectories.

Sometimes it is intended, sometimes not. For example, now looking at typical ignores of a Python project, we surely want to ignore `__pycache__` directories everywhere, so well, we just write `__pycache__`, without any `/` characters.

amotl:
> Thx, but please wait for some other reviewer, more experienced at python.

Thanks for the review. GH-64 would need to go in first, anyway. This PR is stacked on top of it.

testing/ngr.py
```diff
@@ -187,6 +190,55 @@ def test(self) -> None:
...
+    def test(self) -> None:
+        """
+        Invoke software tests.
+        """
+        if self.has_makefile:
+            run_command("make test")
+        elif self.has_pom_xml:
+            run_command(f'mvn test')
+        elif self.has_gradle_files:
+            run_command("./gradlew check")
```
seut, line 237: to run gradle tests, one would issue `./gradlew test`. Or what do I miss?

seut (approved):
lgtm otherwise

testing/ngr.py
amotl: Thanks for spotting. I may have picked this up somewhere else.

amotl: Maybe `test` would invoke the tests, but `check` invokes both the linters, **and** the software tests?

amotl: > The reason _check_ is important is because it can aggregate one or more verification tasks together. This allows us to incorporate multiple tasks into a single step rather than having to execute them one by one.
>
> The Gradle ecosystem contains a number of plugins, such as the [checkstyle](https://docs.gradle.org/current/userguide/checkstyle_plugin.html) plugin, that provide additional functionality to the _check_ task.

-- https://www.baeldung.com/gradle-test-vs-check

seut: Ah ok I see, normally I'd prefer to explicit name tasks one want to run, but as a generic task command this looks feasible.

matriv: Keep in mind that `./gradle check` doesn't work correctly for CrateDB (talking about < 5.5 versions, that use gradle`:
```
~/crate/crate-5.4 (5.4 ✘)⚑✭ ᐅ ./gradlew clean check                                                                                                                                                                                                                 (5.4|✭1)
Starting a Gradle Daemon (subsequent builds will be faster)
....................................................................................................................................................................................................................................................................
> Task :libs:shared:compileJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
.........................................

> Task :libs:guice:compileJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :libs:sql-parser:compileJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :libs:shared:forbiddenApisTest FAILED

> Task :server:compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':libs:shared:forbiddenApisTest'.
> No signatures were added to task; use properties 'signatures', 'bundledSignatures', 'signaturesURLs', and/or 'signaturesFiles' to define those!

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 51s
93 actionable tasks: 75 executed, 18 up-to-date

```

amotl: Thanks. `ngr` will need to handle CrateDB differently then. I will consider this for another iteration. Up until now, it did not need to build CrateDB yet, only auxiliary artefacts.

matriv: yep, it was just FYI, if you ever stumble upon this.

PR #100, opened 2023-10-30
CI: A few adjustments, mostly about scheduling

amotl:
## About

- Dependabot: Add `/framework/{langchain,mlflow}` directories
- Dependabot: Schedule to run weekly
- Reformat `on:` / trigger section
- Schedule all established jobs to run each night. _Only when doing it this way, we will have a chance of getting notified
properly about any changes / regressions. Otherwise, they will go unnoticed, because, well, there is no other activity on the CI job._

### DIFF
--- a/.github/workflows/test-java-jdbc.yml
+++ b/.github/workflows/test-java-jdbc.yml
@@ -1,6 +1,6 @@
 name: Java JDBC
-on:
 
+on:
   pull_request:
     branches: ~
     paths:
--- a/.github/workflows/test-java-jdbc.yml
+++ b/.github/workflows/test-java-jdbc.yml
@@ -17,6 +17,10 @@ on:
   # Allow job to be triggered manually.
   workflow_dispatch:
 
+  # Run job each night after CrateDB nightly has been published.
+  schedule:
+    - cron: '0 3 * * *'
+
 # Cancel in-progress jobs when pushing to the same branch.
 concurrency:
   cancel-in-progress: true
--- a/.github/workflows/test-npgsql.yml
+++ b/.github/workflows/test-npgsql.yml
@@ -19,7 +19,7 @@ on:
 
   # Run job each night after CrateDB nightly has been published.
   schedule:
-    - cron: '0 2 * * *'
+    - cron: '0 3 * * *'
 
 # Cancel in-progress jobs when pushing to the same branch.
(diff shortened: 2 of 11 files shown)


.github/workflows/test-java-jdbc.yml
```diff
@@ -17,6 +17,10 @@ on:
   # Allow job to be triggered manually.
   workflow_dispatch:
 
+  # Run job each night after CrateDB nightly has been published.
+  schedule:
+    - cron: '0 3 * * *'
```
seut, line 22: Afaik the nightly tarball will run at 00:02 and takes ~3min. So this looks too tight, we should run it later. (I see no reason to let it run even an hour later to be safe)

matriv (approved):
nice, thx!

.github/workflows/test-java-jdbc.yml
amotl: Apparently, there is no way to set the time zone anyway. Is it UTC then?

- https://github.com/actions/runner/issues/1177
- https://github.com/actions/runner/issues/1423
- https://github.com/orgs/community/discussions/13454

What do you recommend to use here? Currently, it means "each night at 3 o'clock".

seut (approved):

.github/workflows/test-npgsql.yml
```diff
@@ -21,7 +19,7 @@ on:
 
   # Run job each night after CrateDB nightly has been published.
   schedule:
-    - cron: '0 2 * * *'
+    - cron: '0 3 * * *'
```
amotl, line 22: The job for running the Npgsql tests already has been scheduled to run each night, at `0 2 * * *` cron time. At [^1], we can inspect the scheduled invocations.

The most recent one is indicated to have run at `11 hours ago`. It is [run #6687457412](https://github.com/crate/cratedb-examples/actions/runs/6687457412).

When activating timestamps on the log output, the first timestamp is `Mon, 30 Oct 2023 02:09:30 GMT`. So, it is actually GMT, and we are good to go?


[^1]: https://github.com/crate/cratedb-examples/actions/workflows/test-npgsql.yml

amotl:
It looks like this works well now. Thanks for the positive review!

image
-- https://github.com/crate/cratedb-examples/actions

PR #113, opened 2023-10-31
QA: Add Java JDBC test cases

amotl:
## About

What the title says.

## Why?

- To have more integration tests, and keep them, running against CrateDB nightly.
- To constantly monitor that corresponding test cases don't break, even when
  underlying software versions change. In this case, regular updates to `pom.xml`
  will be driven by GitHub's Dependabot, and Java versions can be added as they
  become available on runners or through corresponding setup recipes.
- Building upon recent improvements here, it is so easy to snap in new folders
  with code snippets, completely independent from each other. Why shouldn't
  it also be re-used for light-weight QA purposes?

## References

- https://github.com/crate/crate-qa/pull/278


by-language/java-qa/src/test/java/io/crate/qa/JdbcTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+@RunWith(JUnit4.class)
+public class JdbcTest {
+
+    @ClassRule
+    public static final CrateTestCluster TEST_CLUSTER = CrateTestCluster
+        .fromURL("https://cdn.crate.io/downloads/releases/nightly/crate-latest.tar.gz")
+        .settings(Map.of("psql.port", 55433))
+        .build();
+    public static final String URL = "jdbc:postgresql://localhost:55433/doc?user=crate";
```
amotl, line 25: It's not strictly required to use `crate-testing` here, because CI/GHA would also provide a running CrateDB instance.

But well, it also tests `crate-testing`, which isn't a total bad thing. The clear drawback is that it currently only works on Linux. If you think we should level up here, let's take it as chance to track it appropriately.

amotl: For example, we could look into whether "Testcontainers for Java" could be used here on behalf of another iteration, to at least give developers on non-Linux platforms a better experience.

by-language/java-qa/src/test/java/io/crate/qa/JdbcTest.java
```diff
@@ -0,0 +1,118 @@
...
+                }
+                int[] results = insert.executeBatch();
+                assertThat(results.length, is(20));
+            }
+
+            stmt.execute("REFRESH TABLE tbl");
+
+            var results = conn.createStatement().executeQuery("SELECT count(*) FROM tbl");
+            assertThat(results.next(), is(true));
+            assertThat(results.getInt(1), is(20));
```
matriv, line 114: Just for completeness you can also add:
```
assertThat(results.next(), is(false));
```

by-language/java-qa/src/test/java/io/crate/qa/JdbcTest.java
matriv: I agree to use testcontainers for this one, but can be done as a followup.

by-language/java-qa/src/test/java/io/crate/qa/JdbcMetaDataTest.java
```diff
@@ -0,0 +1,1238 @@
...
+
+    @Test
+    public void test_autoCommitFailureClosesAllResultSets() throws Exception {
+        try (var conn = DriverManager.getConnection(URL)) {
+            assertThat(conn.getMetaData().autoCommitFailureClosesAllResultSets(), is(false));
+        }
+    }
+
+    @Test
+    public void test_dataDefinitionCausesTransactionCommit_TODO() throws Exception {
```
matriv, line 55: Please remove `TODO`, change the assert to true and add an `@Ignore` annotation, with a description, It will be much more visible, that it's not supported, and easier to find in the future if we ever support commit (Same as `test_getAttributes()`)

by-language/java-qa/src/test/java/io/crate/qa/JdbcMetaDataTest.java
```diff
@@ -0,0 +1,1238 @@
...
+import org.junit.ClassRule;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import io.crate.testing.CrateTestCluster;
+
+@RunWith(JUnit4.class)
+public class JdbcMetaDataTest {
```
matriv, line 24: It's a copy from crate-qa, correct?

matriv (commented):
Thx a lot @amotl, left some comments.

by-language/java-qa/src/test/java/io/crate/qa/JdbcMetaDataTest.java
amotl: This is true. See https://github.com/crate/crate-qa/pull/278.

by-language/java-qa/src/test/java/io/crate/qa/JdbcMetaDataTest.java
amotl: Adjusted with c08fc31dbbbd.

by-language/java-qa/src/test/java/io/crate/qa/JdbcTest.java
amotl: See c08fc31dbbbd.

matriv (approved):
Thank you, LGTM!

PR #128, opened 2023-11-07
ML: Added RAG pipeline with plain SQL usage for Vector Store

ckurze:
- New sample how to use CrateDB's vector store and similarity search with plain SQL.

CC: @marijaselakovic @amotl


amotl (commented):
Thank you very much.

marijaselakovic (approved):
The code works perfectly, thanks @ckurze!

PR #169, opened 2023-11-29
CI: Add Npgsql 8.0.1 to test matrix

amotl:
## About

Npgsql 8.0.0 was released in November 2023. Npgsql 8.0.1 was released in December 2023.

## References
- https://www.npgsql.org/doc/release-notes/8.0.html
- https://github.com/npgsql/npgsql/issues/5503
- https://github.com/crate/crate/issues/15281
- https://github.com/crate/crate-qa/pull/289
- https://github.com/crate/cratedb-examples/pull/169
- [CrateDB 5.5 not working out of the box with Npgsql 8: A PostgreSQL type with the name 'unknown' was not found in the current database info](https://community.cratedb.com/t/cratedb-5-5-not-working-out-of-the-box-with-npgsql-8-a-postgresql-type-with-the-name-unknown-was-not-found-in-the-current-database-info/1689)
- https://github.com/crate/crate/pull/15291
- https://github.com/crate/cratedb-examples/pull/165
- https://github.com/crate/cratedb-examples/pull/195

PR #182, opened 2023-12-04
Npgsql: Stop running tests on .NET framework version 5.x

amotl:
### About

While adding .NET 8.x, we discovered that the testing on .NET 5.x is broken:
> The tests are flawed: They signal "green" status on CI, but actually no tests are compiled and executed.

This patch just removes .NET 5.x from the test matrix.

### References

- https://github.com/crate/cratedb-examples/pull/169#discussion_r1413085186 ff.

### DIFF
--- a/by-language/csharp-npgsql/README.rst
+++ b/by-language/csharp-npgsql/README.rst
@@ -24,7 +24,7 @@ need for a plugin.
 Please note that Npgsql 5 is not supported starting with CrateDB 4.8.4, you
 will need Npgsql 6 or newer.
 
-.NET 5, 6, and 7 are supported, .NET 3.1 and 4.6 may still work.
+.NET 6, 7, and 8 are supported, .NET 3.1, 4.6, and 5.0 may still work.
 
 
 *****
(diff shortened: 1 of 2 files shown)


matriv:
Can you please double check if there is something mentioned in the docs about .NET version support?

amotl:
> Can you please double check if there is something mentioned in the docs about .NET version support?

Are you referring to documentation about CrateDB? This issue here is about the vanilla Npgsql driver. Our canonical page about drivers [^1] does not refer to runtime versions at all.

[^1]: https://cratedb.com/docs/crate/clients-tools/en/latest/connect/

matriv (approved):

matriv:
thx for checking, but sorry, I rushed to approve: we mention versions here: https://github.com/crate/cratedb-examples/blob/main/by-language/csharp-npgsql/README.rst and we need to adjust

amotl:
I've amended the patch by adjusting corresponding version information in the README.

by-language/csharp-npgsql/README.rst
```diff
@@ -37,34 +37,34 @@ To invoke a CrateDB instance for evaluation purposes, run::
 
 Invoke example program::
 
-    dotnet run --framework=net5.0
+    dotnet run --framework=net6.0
```
matriv, line 40: here and below, maybe use `8.0`, the latest? but no strong opinion.

matriv (approved):
thx!

by-language/csharp-npgsql/README.rst
amotl: Sure, why not. I've updated all those references to say `net8.0` instead.

amotl: Now, we need to look into making Npgsql 8 work again.

- GH-169
- https://github.com/crate/crate-qa/pull/289

PR #215, opened 2024-01-04
RAG example with customer data from Twitter

marijaselakovic:
## Checklist

 - [ ] Link to issue this PR refers to (if applicable): 
 - [ ] [CLA](https://crate.io/community/contribute/cla/) is signed

Issue #135, opened 2023-11-08
Use more Markdown


amotl:
## About

Markdown as the lingua franca for many technical writing tasks should be used more, as it is roughly interoperable with, for example, GitHub and Discourse. [^1]

## Conversion from Jupyter Notebooks

Just use [nbconvert](https://pypi.org/project/nbconvert/).

```shell
pip install nbconvert
jupyter nbconvert --to markdown automl_classification_with_pycaret.ipynb
```

## Authoring Jupyter Notebooks

Instead of converting from, Jupyter Notebooks can be written in Markdown itself, see [Notebooks written entirely in Markdown](https://jupyterbook.org/en/stable/file-types/myst-notebooks.html).

> The easiest way to create a MyST notebook is to use [Jupytext](https://jupytext.readthedocs.io/), a tool that allows for two-way conversion between .ipynb and a variety of text files. See also [Notebooks as Markdown](https://jupytext.readthedocs.io/en/latest/formats-markdown.html).

[^1]: Also with HubSpot, when throwing https://github.com/crate-workbench/hubspot-tech-writing into the mix.

amotl:
When looking at [this patch](https://github.com/crate/cratedb-examples/pull/216), it is immediately clear that the slight obfuscation added through the JSON format Jupyter Notebooks are stored in, properly reviewing diffs effectively becomes impossible, because they can't be deciphered well.

I think we should level up here and write Notebooks in Markdown in the future, using, for example, [Jupytext](https://jupytext.readthedocs.io/).

### References
- https://github.com/crate/cratedb-examples/issues/135

/cc @wierdvanderhaar

amotl:
Hi. Thanks for also mentioning that other [page about Quarto, Hex, and Marimo](https://github.com/crate/roadmap/issues/55). Sure enough, those offerings can also be evaluated for publishing content.

Here, we are mostly looking at solutions coming from the Jupyter FOSS communities, pretty close to baseline Jupyter, Binder, and Executable Books technologies.

> Will it still be possible to link directly to these "markdowned" notebooks and open them in Google Colab, for example?

Probably not, we will need to establish a little machinery to also provide/publish the _rendered_ variants when the source file 
is different, i.e. as soon as we will start authoring notebooks like suggested here.

amotl:
> Thanks for also mentioning that other https://github.com/crate/roadmap/issues/55. Those commercial offerings can also be evaluated, but here, we are mostly looking at corresponding FOSS solutions.

[...]

PR #216, opened 2024-01-04
ML/LLM: RAG example copy edits

amotl:
## About

A patch to make GH-215 run to completion, along with a few other adjustments.


amotl:
When looking at this patch, it is immediately clear that the slight obfuscation added through the JSON format Jupyter Notebooks are stored in, properly reviewing diffs effectively becomes impossible, because they can't be deciphered well.

I think we should level up here and write Notebooks in Markdown in the future, using [Jupytext](https://jupytext.readthedocs.io/).

### References
- https://github.com/crate/cratedb-examples/issues/135

/cc @wierdvanderhaar

PR #217, opened 2024-01-07
Apache Superset: Verify connectivity to CrateDB with basic integration tests

amotl:
## About

This patch includes a few basic test cases having conversations with the Apache Superset UI and HTTP API, using a database connection to CrateDB. It is meant to monitor connectivity works, and doesn't break with future releases of both Apache Superset and CrateDB. By running the integration tests each night, the monitoring will be durable.

## Documentation

The README document [Verify Apache Superset with CrateDB](https://github.com/crate/cratedb-examples/blob/b2a01867e7/framework/apache-superset/README.md) included here is a mini tutorial which walks you through the steps needed to setup an instance of Apache Superset on your machine, and connect it to CrateDB. If you don't want to clone any repository to get started, effectively all you need are the [requirements.txt] and [superset_config.py] files.

## Backlog

- [x] Add Apache Superset 3 ~~and Python 3.12~~ to the test matrix on behalf of subsequent patches.
- [ ] The tutorial [Set up an Apache Superset development sandbox with CrateDB] should be updated on a few spots: CrateDB version used should be `latest`, and `CSRF_TOKEN` as well as HTTP session is no longer needed.
- [ ] There should also be a **separate user-oriented tutorial** on the community forum, where all developer-like steps like `git clone` are omitted. Most probably, a user-focused tutorial should be based on Docker Compose for running both Superset and CrateDB, but also not like the upstream documentation [Installing Superset Locally Using Docker Compose] is doing it, because it also uses a `git clone` inside. This new tutorial should be the primary resource to advertise when educating users about this integration, and it should also outline how to connect to CrateDB Cloud.

/cc @karynzv, @marijaselakovic, @ckurze

[Installing Superset Locally Using Docker Compose]: https://superset.apache.org/docs/installation/installing-superset-using-docker-compose/
[requirements.txt]: https://github.com/crate/cratedb-examples/blob/b2a01867e7/framework/apache-superset/requirements.txt
[superset_config.py]: https://github.com/crate/cratedb-examples/blob/b2a01867e/framework/apache-superset/superset_config.py
[Set up an Apache Superset development sandbox with CrateDB]: https://community.cratedb.com/t/set-up-an-apache-superset-development-sandbox-with-cratedb/1163

### DIFF
--- a/framework/apache-superset/conftest.py
+++ b/framework/apache-superset/conftest.py
@@ -104,7 +104,14 @@ def provision_superset(start_superset):
     )
 
     assert response.status_code == 201
-    assert response.json() == {
+    payload = response.json()
+
+    # Superset 3 uses UUIDs to identify resources.
+    # Remove them for comparison purposes.
+    if "uuid" in payload["result"]:
+        del payload["result"]["uuid"]
+
+    assert payload == {
         "id": 1,
         "result": {
             "configuration_method": "sqlalchemy_form",
--- a/framework/apache-superset/conftest.py
+++ b/framework/apache-superset/conftest.py
@@ -1,0 +1,131 @@
+import os
+import shlex
+import shutil
+import subprocess
+import time
+
+import pytest
+import requests
+
+from util import get_auth_headers
+
+
+superset_env = {
+    "FLASK_APP": "superset",
+    "SUPERSET_CONFIG_PATH": "superset_config.py",
+}
+superset_bin = shutil.which("superset")
+
+
+uri_database = "http://localhost:8088/api/v1/database/"
+
+
+# Utility functions.
+
+def invoke_superset(command: str):
+    """
+    Invoke `superset` command.
+    """
+    command = f"{superset_bin} {command}"
+    subprocess.check_call(shlex.split(command), env=superset_env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+
+# Test suite fixtures.
+
+@pytest.fixture(scope="session")
+def fix_greenlet():
+    """
+    Install more recent greenlet, because Playwright installs version 3.0.1, which breaks Superset.
+    """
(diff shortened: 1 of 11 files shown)


framework/apache-superset/conftest.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+# Test suite fixtures.
+
+@pytest.fixture(scope="session")
+def fix_greenlet():
+    """
+    Install more recent greenlet, because Playwright installs version 3.0.1, which breaks Superset.
+    """
+    os.system("pip install --upgrade greenlet")
```
amotl, line 40: Well. Irgendwas ist immer. May need a report or patch on/for Playwright.

amotl:
@vvulf: You recently signaled interest in building something with [Apache ECharts]. On this matter, I'd also recommend to take a look at [Apache Superset], because, well, it's a reasonable container application/framework already for exploring data with visualizations based on ECharts. It can be integrated well through the HTTP API which has been outlined a bit within this patch.

@quodt, @SStorm: [Playwright] hasn't been used much at Crate.io, as far as I know. You also may want to take a look inside this patch, because I think it is the most advanced browser automation tool/interface these days. As with other topics, the devil is in the details of UI testing, but it doesn't take much efforts to quickly whip up a few test cases for UI purposes, thanks to its ergonomic interface and setup procedure, and that it works out of the box on CI/GHA.

[Apache ECharts]: https://echarts.apache.org/
[Apache Superset]: https://superset.apache.org/
[Playwright]: https://github.com/Microsoft/playwright-python

framework/apache-superset/conftest.py
amotl: There is a corresponding patch now.

- https://github.com/microsoft/playwright-python/pull/2227

surister (approved):

hlcianfagna (approved):

PR #218, opened 2024-01-07
Apache Superset: Also verify Superset 3

amotl:
## About

This patch aims to add Apache Superset 3 and Python 3.12 to the CI test matrix, extending GH-217.


amotl:
It looks like Python 3.12 isn't supported yet.

image

-- https://pypi.org/project/apache-superset/

PR #272, opened 2024-02-07
CI: Refactor GHA workflow files into corresponding subdirectories

amotl:
## About

Improve coherency between example snippets and corresponding validation jobs.


amotl:
It looks like GitHub still does not support subfolders within the `.github/workflows` directory. Please upvote.

- https://github.com/orgs/community/discussions/18055

PR #283, opened 2024-02-10
CI: Use Python 3.12 for exercising native Python test layers

amotl:
## About

The "native" tests do not succeed on Python 3.12 yet, cr8 needs a corresponding fix.

```python
  File "/path/to/cratedb-examples/testing/native/python-unittest/test_unittest.py", line 12, in <module>
    from cr8.run_crate import create_node
  File "/path/to/cratedb-examples/.venv312/lib/python3.12/site-packages/cr8/__init__.py", line 1, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

### DIFF
--- a/.github/workflows/testing-native-python.yml
+++ b/.github/workflows/testing-native-python.yml
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'ubuntu-latest', 'macos-latest' ]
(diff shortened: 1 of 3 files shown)


amotl:
There is a corresponding patch now, which will resolve this.

- https://github.com/mfussenegger/cr8/pull/371

PR #291, opened 2024-02-12
TEMP: Probe "testing" demo with updated cratedb-toolkit

amotl:
## About

Aims to verify another patch for cratedb-toolkit.

- https://github.com/crate-workbench/cratedb-toolkit/pull/115


amotl:
Works! 🥳

Issue #313, opened 2024-02-21
CI job trips on Apache Superset 2.x: `AttributeError: module 'flask.json' has no attribute 'JSONEncoder'`


amotl:
## Problem

Invoking `superset db upgrade` fails.

-- https://github.com/crate/cratedb-examples/actions/runs/8005637022/job/21865556761#step:7:290

## History

It failed for the first time on Tue, 20 Feb 2024 03:06:45 GMT. Before, it worked well, and nothing was changed on our ends.

-- https://github.com/crate/cratedb-examples/actions/workflows/application-apache-superset.yml

amotl:
The traceback is:
```python
Traceback (most recent call last):
  File "/path/to/bin/superset", line 5, in <module>
    from superset.cli.main import superset
  File "/path/to/lib/python3.11/site-packages/superset/__init__.py", line 21, in <module>
    from superset.app import create_app
  File "/path/to/lib/python3.11/site-packages/superset/app.py", line 23, in <module>
    from superset.initialization import SupersetAppInitializer
  File "/path/to/lib/python3.11/site-packages/superset/initialization/__init__.py", line 27, in <module>
    from flask_appbuilder import expose, IndexView
  File "/path/to/lib/python3.11/site-packages/flask_appbuilder/__init__.py", line 5, in <module>
    from .api import ModelRestApi  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/path/to/lib/python3.11/site-packages/flask_appbuilder/api/__init__.py", line 38, in <module>
    from ..baseviews import AbstractViewApi
  File "/path/to/lib/python3.11/site-packages/flask_appbuilder/baseviews.py", line 22, in <module>
    from .forms import GeneralModelConverter
  File "/path/to/lib/python3.11/site-packages/flask_appbuilder/forms.py", line 3, in <module>
    from flask_wtf import FlaskForm
  File "/path/to/lib/python3.11/site-packages/flask_wtf/__init__.py", line 4, in <module>
    from .recaptcha import Recaptcha
  File "/path/to/lib/python3.11/site-packages/flask_wtf/recaptcha/__init__.py", line 1, in <module>
    from .fields import RecaptchaField
  File "/path/to/lib/python3.11/site-packages/flask_wtf/recaptcha/fields.py", line 3, in <module>
    from . import widgets
[...]

PR #314, opened 2024-02-21
Chore: Make subprocess.check_call not suppress its output

amotl:
## Problem

It is hard to debug GH-313 when not being able to observe any stderr output from the child process.


amotl:
`superset db upgrade` also fails here. I don't know why.

PR #333, opened 2024-02-26
Autoscaling a CrateDB Cluster out and back

wierdvanderhaar:
- Created a simple Python script that can be used as an example to autoscale a CrateDB Cluster out and back.
- Plus simple readme.


topic/autoscaling/README.md
```diff
@@ -0,0 +1,35 @@
+How to run the autoscale script
+===============================
+
+*****
+About
+*****
+
+This python script can be used to autoscale a CrateDB Cloud Cluster. This script monitors the amount of shards that are on the nodes in the cluster. 
+When this crosses a threshold the cluster will add a node or when the threshold goes under a curtain number the cluster will be scaled-back.
```
hlcianfagna, line 9:
```suggestion
When this crosses a threshold the cluster will add a node and when it goes back down below another threshold the cluster will be scaled-back.
```

topic/autoscaling/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+For more depth and background on this please see the commity post about autoscaling. <ADD LINK>
```
hlcianfagna, line 11:
```suggestion
For more depth and background on this please see the community post about autoscaling. <ADD LINK>
```

topic/autoscaling/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Install script::
+
+    git clone https://github.com/crate/cratedb-examples
+    cd cratedb-examples/topic/autoscaling
+
+
+Change script::
+  Make sure to edit the script before running it. 
+  Update line 7, 10 and 11 with your API credentials, organization id and cluster id.
```
hlcianfagna, line 29:
```suggestion
  Update lines 7, 10, and 11 with your API credentials, organization id and cluster id. 
```

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+            while get_cluster_running_op(cluster_status, cluster_id) != '':
+                cluster_status = get_cluster_status(cluster_id)
+            print('Scaled up successfully!')
+        elif num < (0.5 * max_num_shard):
+            num_nodes = get_cluster_num_nodes(cluster_status, cluster_id)
+            print(f'Start scaling down from {num_nodes + 1} to {num_nodes}')
+            requests.put(f'https://console.cratedb.cloud/api/v2/clusters/{cluster_id}/scale/', json={'product_unit': num_nodes - 1}, auth=auth_data)
+            time.sleep(delay_seconds)
+            while get_cluster_running_op(cluster_status, cluster_id) != '':
+                cluster_status = get_cluster_status(cluster_id)
```
hlcianfagna, line 87: Should we have another `sleep` call inside this `while` loop?

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+    num = num_shards(metrics, cluster_id)  # Calculate average shard count
+    if num is not None:
+        print(f'Current avg number of shards: {num}')
+        if num > (0.8 * max_num_shard):
+            num_nodes = get_cluster_num_nodes(cluster_status, cluster_id)
+            print(f'Start scaling out from {num_nodes + 1} to {num_nodes + 2}')
+            requests.put(f'https://console.cratedb.cloud/api/v2/clusters/{cluster_id}/scale/', json={'product_unit': num_nodes + 1}, auth=auth_data)
+            time.sleep(delay_seconds)
+            while get_cluster_running_op(cluster_status, cluster_id) != '':
+                cluster_status = get_cluster_status(cluster_id)
```
hlcianfagna, line 79: Should we have another sleep call inside this while loop?

topic/autoscaling/README.md
hammerhead: How about rather making those environment variables or arguments to pass to the script? The line numbers can easily change in the future when the script gets adapted.

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+auth_data = HTTPBasicAuth('YOUR API KEY', 'YOUR API SECRET')
+
+# Organization and cluster IDs (to be filled by the user)
+organization_id = 'FILL IN YOUR ORGANIZATION ID'
+cluster_id = 'FILL IN YOUR CLUSTER ID'
+
+# Date format for parsing datetime strings
+date_format = '%Y-%m-%dT%H:%M:%S.%f'
+
+# Function to fetch metrics from CrateDB Cloud for a specific organization
```
hammerhead, line 16: In Python, so-called docstrings go onto the first line of a method: https://peps.python.org/pep-0257/

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+    response = requests.get(api_url, auth=auth_data)
+    return response.text
+
+# Function to calculate the average number of shards per node for a given cluster
+def num_shards(metrics, cluster_id):
+    headers = {'accept': 'plain/text'}
+    api_url = f"https://console.cratedb.cloud/api/v2/organizations/{organization_id}/metrics/prometheus/"
+    response = requests.get(api_url, auth=auth_data, headers=headers)
+    if response.status_code == 200:
+        lines = response.text.split('\n')
```
hammerhead, line 29: How about using the official Prometheus client here to delegate the whole parsing? (https://github.com/prometheus/client_python/tree/master)

Example: https://www.robustperception.io/productive-prometheus-python-parsing/

topic/autoscaling/autoscale.py
wierdvanderhaar: Yeah, that could work. I will add that.

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+            print(f'Start scaling out from {num_nodes + 1} to {num_nodes + 2}')
+            requests.put(f'https://console.cratedb.cloud/api/v2/clusters/{cluster_id}/scale/', json={'product_unit': num_nodes + 1}, auth=auth_data)
+            time.sleep(delay_seconds)
+            while get_cluster_running_op(cluster_status, cluster_id) != '':
+                cluster_status = get_cluster_status(cluster_id)
+            print('Scaled up successfully!')
+        elif num < (0.5 * max_num_shard):
+            num_nodes = get_cluster_num_nodes(cluster_status, cluster_id)
+            print(f'Start scaling down from {num_nodes + 1} to {num_nodes}')
+            requests.put(f'https://console.cratedb.cloud/api/v2/clusters/{cluster_id}/scale/', json={'product_unit': num_nodes - 1}, auth=auth_data)
```
hammerhead, line 85: Would be nice to make the scaling a function as well, as the same code is in this if-branch as well as in the previous one (just with a different parameter). Also would make the main body of the loop a bit easier to read/focus on the execution logic.

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+            if "crate_node" in line and "shard_stats" in line and cluster_id in line and "property=\"total\"" in line:
+                shard_count = float(line.split()[1])
+                total_shard_count += shard_count
+                shard_count_instances += 1
+        if shard_count_instances == 0:
+            return None  # Return None if no shard counts were found
+        else:
+            return total_shard_count / shard_count_instances  # Calculate and return the average
+    else:
+        print(f"Failed to retrieve metrics. Status code: {response.status_code}")
```
hammerhead, line 42: As far as `print` is concerned in general, using it in a logger-like style is somewhat discouraged. For such a small example script it's not a big deal, but on the other hand, using a logger is also straightforward. In the sense of developing a habit of using loggers, I argue for using them:

```python
import logging

logging.getLogger().setLevel('INFO')

logging.info('INFO message')
logging.debug('DEBUG message')
```

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,93 @@
...
+        lines = response.text.split('\n')
+        total_shard_count = 0
+        shard_count_instances = 0
+        for line in lines:
+            if "crate_node" in line and "shard_stats" in line and cluster_id in line and "property=\"total\"" in line:
+                shard_count = float(line.split()[1])
+                total_shard_count += shard_count
+                shard_count_instances += 1
+        if shard_count_instances == 0:
+            return None  # Return None if no shard counts were found
```
hammerhead, line 38: Or just return `0`? Then no special treatment of that case is needed later on and we can get rid of the `if num is not None:` branch.

topic/autoscaling/autoscale.py
```diff
@@ -0,0 +1,94 @@
...
+
+# Function to fetch metrics from CrateDB Cloud for a specific organization
+def get_metrics(organization_id):
+    api_url = f'https://console.cratedb.cloud/api/v2/organizations/{organization_id}/metrics/prometheus/'
+    headers = {'accept': 'plain/text'}
+    response = requests.get(api_url, auth=auth_data)
+    return response.text
+
+# Function to calculate the average number of shards per node for a given cluster
+def num_shards(metrics, cluster_id):
```
hammerhead, line 24: The `metrics` argument isn't used, but metrics are fetched again from the API in this method. Either remove the argument or the second fetching?

topic/autoscaling/autoscale.py
wierdvanderhaar: I changed the methods so that the description is on the first line.

topic/autoscaling/autoscale.py
wierdvanderhaar: Yeah. That would be nice. I had a look, but it raises new challenges, so for now, I will park this and look at it another time.

topic/autoscaling/autoscale.py
wierdvanderhaar: Done that. The the updated version of the script.

topic/autoscaling/autoscale.py
wierdvanderhaar: Changed that.

topic/autoscaling/autoscale.py
amotl: I agree with @hammerhead's idea. However, if the program is functional, it doesn't need to be a blocker, and we can do it on a later iteration.

topic/autoscaling/autoscale.py
wierdvanderhaar: I tried removing this but that keeps failing. So, maybe we can make this change later and leave it as is for now?

topic/autoscaling/autoscale.py
amotl, line 1: I would have a few more comments about the structure of the Python program, and @hammerhead is also looking at linting via black and friends. However, as mentioned elsewhere, I am fine with merging the program if it is functional, adhering to "working code over documentation", and applying any sanding iterations on behalf of subsequent patches.

amotl: The only thing I am asking for is to squash all commits into a single one before merging. Thanks!

topic/autoscaling/autoscale.py
hammerhead: Let's get this sorted out, please. There shouldn't be unused variables or methods in the code.

topic/autoscaling/README.md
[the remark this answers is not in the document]
amotl: Hi @wierdvanderhaar. Please merge GH-334, in order to convert the README to Markdown.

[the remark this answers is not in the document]
amotl: Thanks for merging!

hammerhead (approved):

amotl:
That worked well. Thanks a stack for your contribution. 💯 

image

PR #348, opened 2024-03-05
Anomaly detection notebook

karynzv:
Here is a short notebook on anomaly detection using Pycaret and CrateDB. This was based on this [blog post](https://cratedb.com/blog/introduction-to-time-series-modeling-with-cratedb-machine-learning-time-series-data) by Christian 

## Checklist

 - [ ] Link to issue this PR refers to (if applicable): 
 - [ ] [CLA](https://crate.io/community/contribute/cla/) is signed


topic/timeseries/timeseries-anomaly-detection.ipynb
amotl, line 1: Hi, and thanks for your contribution. Isn't ~50,000 lines pretty large, even for a Jupyter notebook?

ckurze (dismissed):
Thanks for your work on this notebook!

I would suggest that we structure it the same way as EDA and Timeseries Decomposition, so they are almost the same. Nitty-picky here, but would make it easier for users to recognize the same pattern, imho.

The original blog post also visualises the known anomalies first and then compares with the anomaly detection, could this be included as well?

Not sure, but usually the huge amount of lines comes from the plots. Other notebooks (like topic/timeseries/exploratory_data_analysis.ipynb) set the renderer to PNG to avoid it, maybe it solves the large size of the notebook/many lines? 

```
# Plotly plots will be rendered as PNG images, uncomment the following line to crate an interactive chart
plotly.io.renderers.default = 'png'
```

Thank you, great work!

amotl:
> Not sure, but usually the huge amount of lines comes from the plots. Other notebooks (like `topic/timeseries/exploratory_data_analysis.ipynb`) set the renderer to PNG to avoid it, maybe it solves the large size of the notebook/many lines?

Yes, that makes them smaller, at least in terms of "lines of code". Most often, they don't get much larger in terms of size, depending on individual image compression capacities I guess. We will need to inspect the outcome, thanks!

topic/timeseries/timeseries-anomaly-detection.ipynb
```diff
@@ -17,370 +17,19 @@
...
-      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m A new release of pip is available: \u001b[0m\u001b[31;49m23.2.1\u001b[0m\u001b[39;49m -> \u001b[0m\u001b[32;49m24.0\u001b[0m\n",
-      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m To update, run: \u001b[0m\u001b[32;49mpython3.11 -m pip install --upgrade pip\u001b[0m\n",
-      "Note: you may need to restart the kernel to use updated packages.\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
-    "%pip install 'crate[sqlalchemy]' pycaret pandas"
+    "#%pip install 'crate[sqlalchemy]' pycaret pandas"
```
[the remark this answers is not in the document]
karynzv: instead of using `!` I'll replace it for `%` to only install it in the virtual env and not on enforce a global installation, is that ok?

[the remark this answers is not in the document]
marijaselakovic: Sounds good

amotl:
> **Previous size:** 49183 loc · 1.08 MB

**New size:** 732 loc · 222 KB

Excellent, thanks!

topic/timeseries/timeseries-anomaly-detection.ipynb
```diff
@@ -0,0 +1,732 @@
...
+    "                marker=dict(color='red',size=10)))\n",
+    "\n",
+    "fig.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Compared to the initial plot, we can see the model managed to identify the regions were the anomalies happened, but also had a couple of false positives. The model could be further tuned to have a better performance, but that goes beyond this tutorial's goal."
```
amotl, line 701:
```suggestion
    "Compared to the initial plot, we can see the model managed to identify the regions where the anomalies happened, but also had a couple of false positives. The model could be further tuned to have a better performance, but that goes beyond this tutorial's goal."
```

amotl (approved):
Wonderful. Just providing a few more suggestions to use at your own disposal. Thanks again!

marijaselakovic (approved):

topic/timeseries/timeseries-anomaly-detection.ipynb
```diff
@@ -0,0 +1,732 @@
...
+   ],
+   "source": [
+    "query = \"SELECT DATE_BIN('5 min'::INTERVAL, timestamp, 0) AS timestamp, AVG(value) AS avg_value FROM machine_data GROUP BY timestamp ORDER BY timestamp ASC;\"  \n",
+    "\n",
+    "with engine.connect() as conn:\n",
+    "    df = pd.read_sql(sql=sa.text(query), con=conn)\n",
+    "\n",
+    "df['timestamp'] = df['timestamp'].transform(lambda x: datetime.fromtimestamp(x/1000))\n",
+    "\n",
+    "df = df.set_index('timestamp')\n",
```
amotl, line 213: Hi Karyn. Thank you for adjusting this spot to use pandas directly for reading the data. Now, we only have another question out of curiosity, possibly relating to that note:

- https://github.com/crate/crate-python/discussions/589

Is it a common pattern to apply that transformation on the timestamp column, and to define it to be an index on the data frame? For us, it looks like this is the case, that's why we are constantly thinking about improving UX on this matter, even if it would be about getting rid of just two lines of code, or four, depending on how it would be shaped.

So, what do you think about having such an interface for future variants of the same thing?
```python
cratedb = CrateDBAdapter(engine=engine)
df = cratedb.query_df(sql=query, timestamp_columns=["timestamp"])
```

karynzv: Sure, why not? As long ad it is not mandatory to do this transformation, since this is specific to time series.

amotl:
Thanks for merging! 💯

PR #349, opened 2024-03-05
Autoscaling: Refactor code to top-level code environment, and format using `isort`

amotl:
## About
Just some Python code style nit wrt. GH-333.

## Details
For making `autoscale.py` a real program, and good Python citizen, it is
important to consider your file might be considered as a module, and
_imported_. In this case, you don't want to have your main code executed.

https://docs.python.org/3/library/__main__.html

In this case, the program would have caused any sort of test runner,
which scans directory trees for functions to executed, by, well,
importing all modules, to eventually run into an eternal loop.

When putting code into the top-level code environment, as exercised on
behalf of the `__name__ == '__main__'` code guard, it will safely
execute as a Python program, but not when used as a Python module.

In this spirit, a Python file can be both a program and a module at the
same time.

## Miscellaneous
Other than this main change intended to satisfy the patch, bbaa5ceb8e
also formats the code using `isort`.

/cc @hlcianfagna

### DIFF
--- a/topic/autoscaling/autoscale.py
+++ b/topic/autoscaling/autoscale.py
@@ -3,16 +3,16 @@ An example of how to scale your CrateDB Cloud cluster automatically,
 based on its Prometheus metrics endpoint.
 """
 
-import logging
-import time
-import os
 import argparse
-from functools import reduce
+import logging
+import os
+import time
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


hammerhead (approved):
👍

PR #364, opened 2024-03-12
Extend RAG exercise by examples using Jina AI and llamafile

marijaselakovic:
## About
Extend the RAG exercise to use an open-source model through both [Jina AI](https://jina.ai/), or [llamafile](https://github.com/Mozilla-Ocho/llamafile#llamafile).

The LLM at Jina AI is interfaced with through their API, while the llamafile option provides a way to acquire and invoke an LLM on your own premises.


topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -410,11 +460,160 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 42,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from langchain_community.embeddings import JinaEmbeddings\n",
+    "\n",
+    "getenvpass(\"JINA_API_KEY\", prompt=\"Jina API key:\")\n",
+    "embeddings = JinaEmbeddings()"
```
amotl, line 508: Please note: On CI, when not providing the `JINA_API_KEY` to its environment, the test run will exit here using the SKIP outcome, so that it will not fail the software test.

While this is intended for convenience purposes, however, subsequent code will not be tested, and it should be taken into consideration.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -410,11 +460,160 @@
...
+    "### Connect to a local LLM "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "One of the easiest ways to connect to a local LLM is to download one of the models from the Llamafile repository (https://github.com/Mozilla-Ocho/llamafile?tab=readme-ov-file) and run it locally. Upon initiating the llamafile, it not only sets up a web-based user interface for chatting at the local address http://127.0.0.1:8080/, but it also offers an endpoint for chat completions that is compatible with the OpenAI API.\n",
+    "\n",
+    "In this example, we use the Phi-2 model that is about 2GB in size."
```
amotl, line 548: > connect to a local LLM is to download one of the models

It looks like this may need additional infrastructure. Do you think it makes sense to demonstrate that whole topic on behalf of a separate / dedicated notebook? We are only just asking: If you think it is absolutely needed for semantic purposes to have it inside here, we will obey. Otherwise, we are asking to reconsider.

> One of the easiest ways to connect to a local LLM is to download one of the models from the llamafile repository.

What about using [Ollama](https://github.com/ollama/ollama)? Is it a viable alternative to that procedure using [llamafile](https://github.com/Mozilla-Ocho/llamafile)?

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -410,11 +460,160 @@
...
+   "metadata": {},
+   "source": [
+    "### Connect to a local LLM "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "One of the easiest ways to connect to a local LLM is to download one of the models from the Llamafile repository (https://github.com/Mozilla-Ocho/llamafile?tab=readme-ov-file) and run it locally. Upon initiating the llamafile, it not only sets up a web-based user interface for chatting at the local address http://127.0.0.1:8080/, but it also offers an endpoint for chat completions that is compatible with the OpenAI API.\n",
```
amotl, line 546: Just adjusting the repository URL, and making it a clickable link.

```suggestion
    "One of the easiest ways to connect to a local LLM is to download one of the models from the [Llamafile repository](https://github.com/Mozilla-Ocho/llamafile#llamafile) and run it locally. Upon initiating the llamafile, it not only sets up a web-based user interface for chatting at the local address http://127.0.0.1:8080/, but it also offers an endpoint for chat completions that is compatible with the OpenAI API.\n",
```

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
amotl: > What about using [....] to acquire and run an LLM model locally?

Oh wow. That description about llamafile sounds intriguing, indeed. 🍀 

### About llamafile

> llamafile lets you distribute and run LLMs with a single file. ([announcement blog post](https://hacks.mozilla.org/2023/11/introducing-llamafile/))

> Our goal is to make open LLMs much more accessible to both developers and end users. We're doing that by combining [llama.cpp](https://github.com/ggerganov/llama.cpp) with [Cosmopolitan Libc](https://github.com/jart/cosmopolitan) into one framework that collapses all the complexity of LLMs down to a single-file executable (called a "llamafile") that runs locally on most computers, with no installation.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -410,11 +460,160 @@
...
     "chat_completion.choices[0].message.content"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## CrateDB &#129505; OpenSource\n",
+    "\n",
+    "In the second part of this notebook we build a Retrieval-Augmented Generation (RAG) system using CrateDB with Jina AI and the Llama open-source models. \n",
```
amotl, line 485: We'd suggest to split that into a separate notebook, see our comments above. However, if you think it is right that way, we will accept your decision.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
amotl, line 546: At first, when just skimming into the content, it did not become clear to us quickly that this patch actually adds **two** more variants about how to interface with LLMs. One procedure uses the Jina AI API, where a remote LLM is interfaced with, similar to Open AI, and the other procedure uses llamafile for invoking the LLM on your own premises.

Sure enough, it is in the title already:

> using CrateDB with Jina AI and the Llama open-source models.

... but still it wasn't too clear from the very beginning that it is two distinct procedures you are illustrating here.

While it is clear that both new options fit into the "Open Source LLM" box well, we would like to slice their presentations differently, and in a flat manner, so we would actually have **three** dedicated ways how to interface with LLMs, presentation-wise:

- Open AI API (commercial, remote)
- Jina AI API (open-source, remote)
- llamafile (open-source, local)

In this spirit, when using separate notebooks, we would gain dedicated URLs to link to them, may be able to exclude specific ones on the CI/testing procedures selectively, for example when infrastructure or authentication tokens can not be provided there, and would also be able to extend that list of variants more easily in the future, instead of adding even more variants into the same notebook.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -410,11 +460,160 @@
...
+    "\n",
+    "### Configure Jina AI API "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Jina AI embeddings on Hugging Face are a suite of open-source models designed for encoding text data into high-dimensional vectors, which can be used in various natural language processing tasks such as semantic search, text classification, and clustering. These embeddings utilize the BERT architecture and are specially tuned for longer sequence lengths up to 8192 tokens, which is notably higher than the standard BERT model's capacity.\n",
```
amotl, line 494: That sounds sweet. 🎼

amotl (commented):
We love those improvements 💯, and added some suggestions about them. Thank you very much!

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
amotl: Also: Would using LC's `JinaEmbeddings` need other dependencies to be installed, in order to work properly? If so, let's also add them to the `requirements.txt` file?

marijaselakovic: No other dependencies are needed. `JinaEmbeddings` is part of the langchain community module.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
marijaselakovic: There is a overlap in supported models when it comes to Ollama. For instance, it is very convenient to run Phi-2 locally and the Mozilla repository gives a nice set of executable files that have API endpoint compatible with the OpenAI interface. 

For the upcoming presentation, I insist having both examples in a single file. After that we can create a new 'open source' example and play with other models too.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
marijaselakovic: Jina AI is used for calculating embeddings - we can alternatively use HuggingFace or another open-source algorithm. Here there are two procedures:

- Open AI Embedding + LLM models (commerical)
- Jina AI Embeddng + llamafile (open-source)

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
marijaselakovic: Check my comment above. For the presentation purpose it is important to have them together - to avoid redundancy related to data processing step.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
amotl: > For the upcoming presentation, I insist having both examples in a single file. After that we can [refactor] ...

Acknowledged. Thanks for clarifying.

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
amotl: ✅

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
```diff
@@ -320,23 +353,72 @@
...
+      "\u001b[0;31mProgrammingError\u001b[0m: (crate.client.exceptions.ProgrammingError) SQLParseException[vector query dimension: 1536 differs from field dimension: 768]\n[SQL: SELECT embedding.collection_id AS embedding_collection_id, embedding.embedding AS embedding_embedding, embedding.document AS embedding_document, embedding.cmetadata AS embedding_cmetadata, embedding.custom_id AS embedding_custom_id, embedding.uuid AS embedding_uuid, embedding._score AS \"_score\" \nFROM embedding JOIN collection ON embedding.collection_id = collection.uuid \nWHERE embedding.collection_id IN (?) AND knn_match(embedding.embedding, ?, ?) ORDER BY \"_score\" DESC \n LIMIT ?]\n[parameters: ('38b1b477-c407-43c1-b448-5c47dd0677ed', [0.00435982478381371, -0.0070164900711971205, -0.017386690583236015, -0.009653428082896091, -0.034457741030230506, 0.011580167871809374, -0.0488458186 ... (34028 characters truncated) ... 677578292163, -0.013322782385269863, -0.005240995678386175, -0.014730025761513913, -0.01454590048606112, -0.004257898042583784, -0.010015102564799323], 4, 4)]\n(Background on this error at: https://sqlalche.me/e/20/f405)"
+     ]
+    }
```
amotl, line 408: Loads of traceback information here now. Is it intended? Can it be deleted?

amotl (approved):

topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb
marijaselakovic: Another solution will be to delete everything in CrateDB `embedding` table before writing Jira AI embeddings.

amotl: I think one of those options should be applied here, to avoid corresponding confusions, right.

amotl: Thank you. I think it will be good to merge.

Please make sure to squash all your commits into a single one in this case, because individual ones are noisy, also partly removing noise introduced by previous ones, so they don't have much value to be forwarded into mainline individually. Thanks!

PR #366, opened 2024-03-13
Fix dependency installation for `topic/machine-learning/automl`

hammerhead:

Google Colab currently fails running `pip install`. `pycaret[analysis]` pulls in `pyyaml==5.3.1` (from 2020!), which no longer installs on Google Colab:

```
Collecting pyyaml==5.3.1 (from pycaret[analysis,models,parallel,test,tuner]==3.3.0->-r https://raw.githubusercontent.com/crate/cratedb-examples/main/topic/machine-learning/automl/requirements.txt (line 5))
  Using cached PyYAML-5.3.1.tar.gz (269 kB)
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  Preparing metadata (setup.py) ... error
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

I couldn't find any explicit documentation from PyCaret on what the `analysis` extra does. On [master](https://github.com/pycaret/pycaret/blob/master/requirements-optional.txt), PyCaret already updated PyYAML to 5.4.1 (from 2021 🙄). Can anyone confirm if the extra is actually used?

There was also the `tuner` extra used which doesn't exist (any more) as well as a few direct dependencies I could not find any usages of.

## Checklist

 - [X] Link to issue this PR refers to (if applicable): 
 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/topic/machine-learning/automl/requirements.txt
+++ b/topic/machine-learning/automl/requirements.txt
@@ -5,7 +5,6 @@ plotly<5.20
 pycaret[models,parallel,test]==3.3.0
 pydantic<2
 python-dotenv<2
-tqdm<5
 
 # Development.
 # mlflow-cratedb @ git+https://github.com/crate-workbench/mlflow-cratedb.git@main
--- a/topic/machine-learning/automl/requirements.txt
+++ b/topic/machine-learning/automl/requirements.txt
@@ -2,7 +2,7 @@
 crate[sqlalchemy]
 mlflow-cratedb==2.10.2
 plotly<5.20
-pycaret[analysis,models,parallel,test]==3.3.0
+pycaret[models,parallel,test]==3.3.0
 pydantic<2
 python-dotenv<2
 tqdm<5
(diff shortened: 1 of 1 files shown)


amotl (approved):
Thank you for resolving an issue with dependencies we have not been aware of. We fully trust your assessment on this topic, as we also only added them on a best-effort basis.

While we can't trust CI fully, because it just runs the notebooks to completion, without actually validating them, we think it is at least a soft indicator that CI succeeds, that the change will not cause any other havoc, and demonstrates the pieces also work in a non-Colab environment, right?

hammerhead:
Right, I didn't have the chance yet to run through both of the Jupyter Notebooks. That should still be done prior to merging this, ensuring it's not failing on any missing dependencies with this change.

amotl:
We think it will be good, because CI usually tests **all the notebooks** within the corresponding folder. Any missing dependency would be expected to raise a runtime error there, so it should not go unnoticed. Coming from this assessment, we think it is safe to merge this patch.

Issue #424, opened 2024-04-18
RAG: Problems resolving Dask dependency on Google Colab


amotl:
## Problem

@hammerhead reported a flaw with the [cratedb_rag_customer_support_langchain.ipynb](https://colab.research.google.com/github/crate/cratedb-examples/blob/main/topic/machine-learning/llm-langchain/cratedb_rag_customer_support_langchain.ipynb) Notebook when invoked on Google Colab.

Dependency resolution around Dask fails, bzw. takes ages to complete, if at all.

```
Collecting dask (from fsspec[abfs,dask,gcs,git,github,http,s3,smb]<2024.3->pueblo[cli,fileio,nlp]>=0.0.7->-r https://raw.githubusercontent.com/crate/cratedb-examples/main/topic/machine-learning/llm-langchain/requirements.txt (line 9))
  Downloading dask-2022.5.2-py3-none-any.whl (1.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.1/1.1 MB 65.4 MB/s eta 0:00:00
Collecting distributed (from fsspec[abfs,dask,gcs,git,github,http,s3,smb]<2024.3->pueblo[cli,fileio,nlp]>=0.0.7->-r https://raw.githubusercontent.com/crate/cratedb-examples/main/topic/machine-learning/llm-langchain/requirements.txt (line 9))
  Downloading distributed-2022.5.1-py3-none-any.whl (871 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 871.6/871.6 kB 54.8 MB/s eta 0:00:00
Collecting dask (from fsspec[abfs,dask,gcs,git,github,http,s3,smb]<2024.3->pueblo[cli,fileio,nlp]>=0.0.7->-r https://raw.githubusercontent.com/crate/cratedb-examples/main/topic/machine-learning/llm-langchain/requirements.txt (line 9))
  Downloading dask-2022.5.1-py3-none-any.whl (1.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.1/1.1 MB 57.8 MB/s eta 0:00:00
[...]

PR #428, opened 2024-04-18
LLM: Get rid of dependency to `pueblo[fileio]`

amotl:
## Problem

`pueblo[fileio]`, being a bit heavy, caused troubles on Google Colab, as reported by @hammerhead.

- GH-424

## Solution

This patch gets rid of this dependency completely.


hammerhead (approved):
Thanks!

PR #442, opened 2024-05-03
TypeORM: ID column to string to support UUID

hlcianfagna:
## About
Adjust example about that primary autogenerated keys with CrateDB are UUIDs, stored as strings.

## References
This example uses those other dialect patches needed to make TypeORM compatible with CrateDB.
- https://github.com/crate-workbench/typeorm/pull/1
- https://github.com/crate-workbench/typeorm/pull/2

It is also stacked upon this one.
- GH-187


amotl (approved):
Thanks!

Issue #480, opened 2024-06-18
ML/LLM: `cratedb_rag_customer_support_langchain.ipynb` croaks with `ImportError: cannot import name 'patch_inspector' from 'cratedb_toolkit.sqlalchemy.patch'`


amotl:
@wierdvanderhaar reported another problem. Thank you.

```python
WARNING: langchain 0.2.3 does not provide the extra 'cratedb'
WARNING: langchain 0.2.3 does not provide the extra 'openai'

ImportError                               Traceback (most recent call last)
<ipython-input-2-451439e6f556> in <cell line: 11>()
      9 from langchain_openai import OpenAIEmbeddings
     10 from langchain_community.document_loaders import CSVLoader
---> 11 from langchain_community.vectorstores import CrateDBVectorSearch
     12 
     13 warnings.filterwarnings('ignore')

3 frames
/usr/local/lib/python3.10/dist-packages/langchain_community/vectorstores/cratedb/base.py in <module>
     15 
     16 import sqlalchemy
---> 17 from cratedb_toolkit.sqlalchemy.patch import patch_inspector
     18 from cratedb_toolkit.sqlalchemy.polyfill import (
     19     refresh_table,

ImportError: cannot import name 'patch_inspector' from 'cratedb_toolkit.sqlalchemy.patch' (/usr/local/lib/python3.10/dist-packages/cratedb_toolkit/sqlalchemy/patch.py)
```

amotl:
Apparent quick fix:
```
pip install cratedb-toolkit==0.0.12
```

amotl:
Another improvement: ff48b76f584.

amotl:
Apparently, it worked only on Google Colab. On CI, installing cratedb-toolkit 0.0.12 clearly shows it is not compatible.

-- https://github.com/crate/cratedb-examples/actions/runs/9565093245/job/26367271413?pr=482#step:6:480

amotl:
I will revert ef3c38df94 again, and would like to ask you to re-validate your environment, possibly wiping it completely.

amotl:
```python
ImportError: cannot import name 'patch_inspector' from 'cratedb_toolkit.sqlalchemy.patch'
```
Now also on CI: https://github.com/crate/cratedb-examples/actions/runs/9565423703/job/26368328266?pr=482#step:6:803

Excellent, there must be a dependency hiccup then, which can deterministically be fixed.

amotl:
[...]

Issue #479, opened 2024-06-18
timeseries-queries-and-visualization.ipynb - Loading initial data doesnt return output


wierdvanderhaar:
The loading of data doesnt return the output that is described. 

`The result contains information about the successfully written rows and potential errors that might have occurred. The output is expected to look roughly like this:

[({'id': '<SOME_ID>', 'name': '<SOME_NAME>'},
  'https://github.com/crate/cratedb-datasets/raw/main/cloud-tutorials/data_weather.csv.gz', 70000, 0, {} )]
The response indicates that 70,000 records have been successfully loaded and that no errors happened.'

My suggestion would be to fully get rid of this piece of text as this doesn't match the reality.

amotl:
Thanks. 62cf293ae87 will get rid of this cell.

Issue #478, opened 2024-06-18
Time Series: Incompatibility between Google Colab and pandas 2


wierdvanderhaar:
When running timeseries-queries-and-visualization.ipynb in Colab you get the following error when installing the needed pip packages. 

``` python
!pip install --upgrade kaleido 'pandas>=2' plotly sqlalchemy-cratedb
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf-cu12 24.4.1 requires pandas<2.2.2dev0,>=2.0, but you have pandas 2.2.2 which is incompatible.
google-colab 1.0.0 requires pandas==2.0.3, but you have pandas 2.2.2 which is incompatible.
````

This can be fixed by running:
``` python
!pip install --upgrade kaleido 'pandas==2.0.3' plotly sqlalchemy-cratedb
```

amotl:
Thanks. Will be fixed with fa5edbd10b46c2.

amotl:
Google Colab's changelog reports that on [2024-05-13](https://colab.research.google.com/notebooks/relnotes.ipynb#scrollTo=fea5a3f2), the `cudf` package has been included.

> Python package inclusions
> - cudf 24.4.1

This apparently wants to pull in pandas=>2.0, [...]
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
cudf-cu12 24.4.1 requires pandas<2.2.2dev0,>=2.0, but you have pandas 1.5.3 which is incompatible.
cudf-cu12 24.4.1 requires pyarrow<15.0.0a0,>=14.0.1, but you have pyarrow 16.1.0 which is incompatible.
google-colab 1.0.0 requires pandas==2.0.3, but you have pandas 1.5.3 which is incompatible.
google-colab 1.0.0 requires requests==2.31.0, but you have requests 2.32.3 which is incompatible.
ibis-framework 8.0.0 requires pyarrow<16,>=2, but you have pyarrow 16.1.0 which is incompatible.
```

[...] but our "Time Series" stack is not up-to-date to be able to use it.

- GH-387
- GH-430

amotl:
The situation will be improved with:

- GH-430

Upgrading two other notebooks has been postponed:

- GH-493

PR #482, opened 2024-06-18
Fix notebooks about time series and ML/LLM

amotl:
## About

Bundles fixes and improvements reported and/or suggested by @wierdvanderhaar. Thank you.

- GH-478
- GH-479
- GH-480


wierdvanderhaar (approved):
Looks good to me.

PR #483, opened 2024-06-18
Synchronize dependencies by using `requirements.txt` across the board

amotl:
## Problem
Diverging sets of dependencies are a constant struggle for people vs.
machines, either aiming to use the notebooks, or to validate them,
respectively.

## Proposal
The idea is to improve the situation by referring to the canonical
`requirements.txt` file from everywhere. In this spirit, dependency
updates can be driven well by Dependabot, and relevant software
artefacts should be synchronized better.

## Trivia
- By using absolute URLs to the `requirements.txt` files, the relevant `pip install` invocation should also work well on runtime environments like Google Colab, which need them.
- By using the `--upgrade` / `-U` option for `pip install`, let's try to ensure dependencies are always up-to-date.

### DIFF
--- a/topic/machine-learning/automl/README.md
+++ b/topic/machine-learning/automl/README.md
@@ -91,7 +91,7 @@ Alternatively, you can validate all resources in this folder by invoking a
 test runner program on the top-level folder of this repository. This is the
 same code path the CI jobs are taking.
 ```shell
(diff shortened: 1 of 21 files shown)


wierdvanderhaar (approved):
Looks good to me.

PR #484, opened 2024-06-18
Dependencies: Use fixed version constraints for more volatile packages

amotl:
## Problem
When refactoring, even diligently, things occasionally go south. Apologies.

- GH-478
- GH-480
- GH-481

## Proposal
Provide better protection against accidental havocs, by nailing packages with more volatility using exact version constraints.


wierdvanderhaar (approved):
Looks steady to me.

PR #485, opened 2024-06-18
CI: Let Dependabot probe dependency updates on a daily basis

amotl:
## About
Intending to improve the situation with recent dependency havocs.

## Details
For more volatile packages, this will increase churn. For all others, and in general, it will report sad dependency situations earlier than before.

## References
- GH-484
- GH-483


wierdvanderhaar (approved):
Daily seems a smart interval for these notebooks and their checks.

PR #430, opened 2024-04-19
Time Series: Update to pandas 2.0 and SQLAlchemy 2.0

amotl:
## About
For the time series notebooks, update pandas and SQLAlchemy to use more recent versions.

## Details
An alternative to GH-400, where anyone can collaborate on the same branch. This makes sense if it is a long running one, because, for example, when it needs contributions from more people while we go.

## References
- GH-379
- GH-383
- GH-477
- GH-476

Issue #387, opened 2024-03-19
Time Series: Modernize notebooks to use recent versions of pandas and SQLAlchemy


amotl:
## Problem
Even if only just conceived, the new notebooks are immediately outdated already, using legacy technologies like pandas 1.x and SQLAlchemy 1.x.

- https://github.com/crate/cratedb-examples/pull/379
- https://github.com/crate/cratedb-examples/pull/383

## Solution
Provide updates to use more recent / modern versions of both pandas and SQLAlchemy.

amotl:
@wierdvanderhaar reported that we dearly need this update.

- GH-430

Does anyone have any capacity to fix this? /cc @surister, @wierdvanderhaar, @ckurze, @hammerhead

PR #492, opened 2024-06-18
Time Series: Fix dependencies, explicitly using `pandas==2.0.*`

amotl:
## About
On resolving dependency issues with the time series notebook(s), @wierdvanderhaar suggested to add these dependencies. They seem to specifically be needed to satisfy Google Colab?


amotl:
Using pandas 2.x is not an option, at least not for the `exploratory_data_analysis.ipynb` notebook in the same folder.

- GH-387
- GH-430

PR #494, opened 2024-06-18
PHP/AMPHP: Update to amphp/amp v3 by switching to `Revolt\EventLoop`

amotl:
## About
- Updates to amphp/amp 3.x. -- https://amphp.org/amp
- Includes and extends GH-487.

## Trivia
- Remove support for PHP 7. Add support for PHP 8.3.

### DIFF
--- a/.github/workflows/lang-php-amphp.yml
+++ b/.github/workflows/lang-php-amphp.yml
@@ -37,7 +37,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'ubuntu-latest' ]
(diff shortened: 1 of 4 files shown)


surister (approved):

surister:
just a candid question, do we know who uses cratedb + php?

amotl:
> do we know who uses cratedb + php?

There are a few customers or prospects here and there. They even use PHP together with MQTT, iirc. Please correct me if I'm wrong. /cc @karynzv, @hlcianfagna, @hammerhead

PR #509, opened 2024-07-03
Kafka/Flink: Fix DDL statement

wierdvanderhaar:
docker-compose is a separate executable nowadays.  
The ingest from Flink fails when you use the old create table statement because only location and current are ingested. With these generated columns, we are sure that the ingests are successful. 

## Checklist

 - [ ] Link to issue this PR refers to (if applicable): 
 - [ ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/application/apache-kafka-flink-streaming/README.md
+++ b/application/apache-kafka-flink-streaming/README.md
@@ -40,7 +40,7 @@ into your local `.env` file.
 ### Run the docker compose (and build the images)
 
 ```
-docker compose up -d --build
+docker-compose up -d --build
 ```
 
 ### Stop the docker compose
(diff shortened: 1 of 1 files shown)


application/apache-kafka-flink-streaming/README.md
```diff
@@ -40,7 +40,7 @@ into your local `.env` file.
 ### Run the docker compose (and build the images)
 
 ```
-docker compose up -d --build
+docker-compose up -d --build
```
amotl, line 43: That would use an older Docker Compose I guess. The new one is part of the Docker command itself.

-- https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

amotl: Let's stick with the previous incantation, when possible?

amotl (approved):

amotl:
Hi. Thanks for your fix. For the sake of clarity, please use a better commit message.

surister (approved):

PR #695, opened 2024-10-24
Npgsql: Adjust .NET support and version numbers on CI

amotl:
## About
- Remove support for .NET 6 and Npgsql 6.
- Add support for .NET 9.
- Update CI to use Npgsql versions 7.0.8 and 8.0.5.

## References
- GH-675

### DIFF
--- a/.github/workflows/lang-npgsql.yml
+++ b/.github/workflows/lang-npgsql.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         os: [ 'ubuntu-latest' ]
         dotnet-version: [ '7.0.x', '8.0.x', '9.0.x' ]
-        npgsql-version: [ '7.0.7', '8.0.3' ]
+        npgsql-version: [ '7.0.8', '8.0.5' ]
         cratedb-version: [ 'nightly' ]
 
     # https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers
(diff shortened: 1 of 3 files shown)


.github/workflows/lang-npgsql.yml
```diff
@@ -42,8 +42,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'ubuntu-latest' ]
-        dotnet-version: [ '6.0.x', '7.0.x', '8.0.x' ]
-        npgsql-version: [ '6.0.11', '7.0.7', '8.0.3' ]
+        dotnet-version: [ '7.0.x', '8.0.x', '9.0.x' ]
```
kneth, line 45: .NET 7 is no longer maintained while .NET 6 is (until November 2024). Would it make sense to wait for .NET 9 to be released (currently only released as RC)?

kneth (approved):
We can definitely merged now but I expect that we need to adjust version numbers again in late November.

surister (approved):

.github/workflows/lang-npgsql.yml
amotl: ## .NET 9

> Would it make sense to wait for .NET 9 to be released (currently only released as RC)?

We are not doing any super strong promises here. I am always happy to unlock CI slots as soon as artefacts of major releases become available, even as RC versions, so our assets can be tested on them from the very beginning.

If you don't have any objections agains that philosophy, let's embrace it?

kneth: > Do you think we should remove it from the CI matrix?

As long as it doesn't hurt us, we can keep it.

> don't want to terminate validation of .NET 6 too early

.NET 6 is gone next month anyway, so let's remove it instead of spending time on figuring out why it failed. It is unlikely any user will complain.

PR #706, opened 2024-11-01
Python DB API: Add basic example about i/o with vector embeddings

amotl:
## About
We needed to verify that querying CrateDB using basic operations with »`vector_similarity()` + parameters« works well. It does.

## Trivia
While at it, the patch uses the chance to restructure the examples to a "by topic" layout instead of a "by operation" one.


amotl:
Within a basic example scenario, everything works as expected. Over here, we are hitting an edge case:

- https://github.com/crate/crate/issues/16912

kneth (approved):

PR #750, opened 2024-11-23
Metabase: Add test harness for validating Metabase against CrateDB

amotl:
## About
A basic test case that reads CrateDB's `sys.summit` table through Metabase, after connecting CrateDB as a PostgreSQL database.

## Details
In order to prove it worked in general, this recipe uses Metabase v0.48.3, while it will start failing with v0.48.4.

## References
- https://github.com/crate/crate-clients-tools/issues/153

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -90,6 +90,16 @@ updates:
     schedule:
       interval: "daily"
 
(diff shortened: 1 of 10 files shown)


kneth (approved):
LGTM

PR #762, opened 2024-12-05
Azure Function Python v2

karynzv:
This is an example of an enrichment process function which consumes events from Event Hub, processes them and insert into CrateDB.

## Checklist


amotl (commented):
Thanks. Didn't do a thorough review, but added one little suggestion. I'd also like to change the folder name and add a little software test case, but this can also happen after merging.

hammerhead:
Thanks for making it work with the v2 model!

For this to be an example of basic interoperability between CrateDB/Azure Functions, I think we can trim this down quite a bit. The code originates from a specific implementation project that has all sorts of complexity and corner cases. Which I feel we can skip here, trying to make this a concise example showing how *easy* it is to make the integration work.

Examples:
* Duplicate storage `raw`/`reading`: Would it be our default recommendation to store both pre/post enrichment payloads? We could eliminate the `raw` table, try to insert directly into `reading`, and if it fails, iterate through the batch row-by-row and insert into `error` if needed. That would eliminate the whole `ValueCache` complexity, looking up rows by `trace_id` and move them from one list to another.
* Speaking of `trace_id`, that could then also go away if we don't need to correlate rows between `raw` and `reading` anymore.
* Data model: I noticed `reading` doesn't use `OBJECT` but has the JSON attributes transformed to top-level columns. Using `OBJECT(DYNAMIC)` is one of the strengths of CrateDB, I suggest using it. That could also reduce the complexity of enrichment if we just pass the original payload through as-is. Or make a very lightweight example of how one can do simple transformations, such as renaming a JSON attribute before insertion, if needed.

topic/Azure function - event hub triggered/Azure Function/crate_writer.py
```diff
@@ -0,0 +1,127 @@
+import time
+import logging
+from crate import client
+from crate.client.exceptions import ProgrammingError, IntegrityError
+
+
+class CrateWriter:
```
hammerhead, line 7:
```suggestion
class CrateDBWriter:
```
Just a small naming detail for Crate -> CrateDB

topic/Azure function - event hub triggered/Azure Function/value_cache.py
```diff
@@ -0,0 +1,32 @@
+import logging
+
+class ValueCache:
+    """
+    ValueCache class is used to structure enriched data for insert.
+    """
+
+    def __init__(self):
+        self.errors = []
+        self.errors_empty = []
```
hammerhead, line 10: This `errors_empty` we can also eliminate? At least I don't see any further usage of it, besides appending a few lines down.

topic/Azure function - event hub triggered/Azure Function/enrichment.py
```diff
@@ -0,0 +1,71 @@
...
+            f"-- payload: {raw_payload}"
+        )
+        value_cache.add_error(raw_payload, str(e), type(e).__name__)
+    
+def transform_payload(event):
+    #remove empty keys
+    event = remove_empty_keys(event)
+    #change column names
+    event = rename_keys(event)
+    #check for sensor_id, timestamp, location keys
```
hammerhead, line 52: Can you run `black` please on the code to align formatting?

hammerhead (approved):

amotl (approved):
Thank you. Two more suggestions from my side, feel free to amend.

- Folder name: `topic/serverless/azure-eventhub`.
- Squash all commits into a single one.

PR #771, opened 2024-12-11
Npgsql: Fix test cases to use async, stop relevant warnings

amotl:
What the title says.


kneth (approved):
LGTM

It looks like [xUnit1031](https://xunit.net/xunit.analyzers/rules/xUnit1031) was introduced long time ago so it is surprising that it didn't surface until the Ubuntu GHA runner was changed.

amotl:
> It looks like [xUnit1031](https://xunit.net/xunit.analyzers/rules/xUnit1031) was introduced long time ago so it is surprising that it didn't surface until the Ubuntu GHA runner was changed.

CI was accompanied by relevant warnings for a long time already. There is nothing related to the runner change here.

PR #772, opened 2024-12-12
Npgsql: Add test validating (almost) all data types of CrateDB

amotl:
## About
What the title says.


by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -0,0 +1,188 @@
...
+                cmd.Parameters.AddWithValue("timestamp_tz", "1970-01-02T00:00:00+01:00");
+                cmd.Parameters.AddWithValue("timestamp_notz", "1970-01-02T00:00:00");
+                cmd.Parameters.AddWithValue("ip", "127.0.0.1");
+                cmd.Parameters.AddWithValue("array", new List<string>{"foo", "bar"});
+                // FIXME: System.NotSupportedException: Cannot resolve 'hstore' to a fully qualified datatype name. The datatype was not found in the current database info.
+                // cmd.Parameters.AddWithValue("object", new Dictionary<string, string>(){{"foo", "bar"}});
+                cmd.Parameters.AddWithValue("geopoint", new List<double>{85.43, 66.23});
+                // FIXME: Npgsql.PostgresException : XX000: line 38:9: no viable alternative at input 'VALUES
+                // cmd.Parameters.AddWithValue("geoshape", "POLYGON ((5 5, 10 5, 10 10, 5 10, 5 5))");
+                cmd.Parameters.AddWithValue("float_vector", new List<double> {1.1, 2.2, 3.3});
```
amotl, line 163: This is the spot that best conveys the capabilities of the vanilla Npgsql driver together with CrateDB, where we currently can't send values of type OBJECT or GEOSHAPE using parameterized INSERT operations, at least not in the way presented here.

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -0,0 +1,188 @@
...
+                cmd.Parameters.AddWithValue("bit", "01010101");
+                cmd.Parameters.AddWithValue("bool", true);
+                cmd.Parameters.AddWithValue("text", "foobar");
+                cmd.Parameters.AddWithValue("char", "foo");
+                cmd.Parameters.AddWithValue("timestamp_tz", "1970-01-02T00:00:00+01:00");
+                cmd.Parameters.AddWithValue("timestamp_notz", "1970-01-02T00:00:00");
+                cmd.Parameters.AddWithValue("ip", "127.0.0.1");
+                cmd.Parameters.AddWithValue("array", new List<string>{"foo", "bar"});
+                // FIXME: System.NotSupportedException: Cannot resolve 'hstore' to a fully qualified datatype name. The datatype was not found in the current database info.
+                // cmd.Parameters.AddWithValue("object", new Dictionary<string, string>(){{"foo", "bar"}});
```
amotl, line 159: For making CrateDB's OBJECT data types work with vanilla adapters/drivers that have been designed for talking to PostgreSQL, there is the opportunity to converge early PostgreSQL's HSTORE feature, or recent PostgreSQL's JSON(B) feature into relevant support for CrateDB's OBJECT, in fact reasonably transparently, by having extension code in the right place.

A recent adapter (not released yet) for Python/SQLAlchemy is inside here.
- https://github.com/crate/sqlalchemy-cratedb/pull/27

amotl: Adapters for HSTORE are part of our custom drivers for Java, .NET, and others, from the very beginning of CrateDB becoming PostgreSQL compatible, because there have been enough insights that it needs special extension code to fundamental drivers always and anyway, in order to support CrateDB's special features.

However, it looks like attention in this area has **decreased** after the focus shifted to **increase** PostgreSQL compatibility of CrateDB itself. In these times, it seems to have been forgotten that there will always be special features that CrateDB needs support for, and therefore, driver support will always be needed.
- https://github.com/search?q=org%3Acrate%20hstore&type=code

This idea went so far that we event flagged the JDBC driver as "legacy", which led to a lot of downstream confusion. We are gradually reverting that idea, and made a start over here.

- https://github.com/crate/crate-jdbc/pull/439
- https://github.com/crate/crate-jdbc/pull/440

Sure enough, in a perfect world, we would upstream our special needs into the canonical drivers, but currently, we are not even close to that.

amotl: In the past, the company was pretty active in this regard, 
- https://github.com/npgsql/npgsql/issues/1486
- https://github.com/npgsql/npgsql/commit/bb0f5b54f789fb58a80fd339c29dc806c91e8a28
- https://github.com/npgsql/npgsql/pull/1660
- https://github.com/npgsql/npgsql/pull/1938
- https://github.com/npgsql/npgsql/pull/2202
- https://github.com/npgsql/npgsql/pull/2318

The outcome of all of that in fact was the [Npgsql **Plugin** for CrateDB](https://cratedb.com/docs/npgsql/), which exactly [adds support for CrateDB's special data types](https://cratedb.com/docs/npgsql/en/latest/appendices/data-types.html), at least for OBJECTs and ARRAYs of OBJECTs, which are the most important container types, and one of the killer features of CrateDB anyway.

However, maintenance on this plugin dried out, possibly due to misunderstandings, but certainly mostly due to resource drain.
- https://github.com/crate/crate-npgsql/issues/4

The reality in 2024 is, that this plugin, [crate-npgsql](https://github.com/crate/crate-npgsql), depends on a version of canonical Npgsql from six years ago, exactly like the CrateDB JDBC driver is doing it. I dearly think we need to resurrect this, in order to provide premium support for CrateDB's users, including support for special data types, instead of referring them to the standard PostgreSQL driver, which currently just does not provide those features that are actually important for users using CrateDB.

amotl: In order to improve the situation, I started to track this internally.
- https://github.com/crate/product/issues/26

I think a reasonable plan, after resurrecting maintenance of the CrateDB Npgsql Plugin, is to work through **all data types already provided by Npgsql**, and make them work reasonably well together with CrateDB.
- BASIC: https://www.npgsql.org/doc/types/basic.html
- OBJECT/ARRAY: https://www.npgsql.org/efcore/mapping/json.html
- GEO: https://www.npgsql.org/doc/types/geojson.html
- VECTOR: https://github.com/pgvector/pgvector-dotnet, https://github.com/xakpc/DotnetPrompt.Npgsql.PgVector

/cc @kneth, @simonprickett

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -0,0 +1,188 @@
...
+                cmd.Parameters.AddWithValue("geopoint", new List<double>{85.43, 66.23});
+                // FIXME: Npgsql.PostgresException : XX000: line 38:9: no viable alternative at input 'VALUES
+                // cmd.Parameters.AddWithValue("geoshape", "POLYGON ((5 5, 10 5, 10 10, 5 10, 5 5))");
```
amotl, line 162: Maybe GEOSHAPE can be made work easily. I haven't yet investigated what is behind a hiccup in this case. The primary reason of this patch is to evaluate and educate, and to set the stage for relevant insights and discussions.

amotl (commented):
Hi. I am adding some insights about this patch at the spots where they are relevant.

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -0,0 +1,188 @@
...
+                cmd.Parameters.AddWithValue("bigint", 42);
+                cmd.Parameters.AddWithValue("float", 42.42);
+                cmd.Parameters.AddWithValue("double", 42.42);
+                cmd.Parameters.AddWithValue("decimal", 42.42);
+                cmd.Parameters.AddWithValue("bit", "01010101");
+                cmd.Parameters.AddWithValue("bool", true);
+                cmd.Parameters.AddWithValue("text", "foobar");
+                cmd.Parameters.AddWithValue("char", "foo");
+                cmd.Parameters.AddWithValue("timestamp_tz", "1970-01-02T00:00:00+01:00");
+                cmd.Parameters.AddWithValue("timestamp_notz", "1970-01-02T00:00:00");
```
amotl, line 155: I am not sure about those, because apparently, they work in this context, but it is also just a very basic test anyway.
Let me reference this report here, to have a complete list of evalations here, and to reduce the surface of unknown unknowns.

- https://github.com/crate/crate-npgsql/issues/67

by-language/csharp-npgsql/DemoTypes.cs
amotl: Basic support for OBJECT types (marshalled using STRING) is validated per 23911c579ce97.

by-language/csharp-npgsql/DemoTypes.cs
amotl: Basic support for GEO_SHAPE types (marshalled using STRING) is now validated per 131972aef16.

by-language/csharp-npgsql/tests/DemoProgramTest.cs
```diff
@@ -79,6 +81,91 @@ public async Task TestUnnestExample()
...
+            var dt = await task.WaitAsync(TimeSpan.FromSeconds(0.5));
+
+            // Check results.
+            var row = dt.Rows[0];
+
+            // Numeric types
+            Assert.Equal(DBNull.Value, row["null_integer"]);
+            Assert.Equal(42, row["integer"]);
+            Assert.Equal((Int64) 42, row["bigint"]);
+            //Assert.Equal(42.42, (float) row["float"], 0.01);
```
kneth, line 101: Why are these asserts disabled?

kneth (commented):
LGTM

by-language/csharp-npgsql/tests/DemoProgramTest.cs
amotl: It was probably just an oversight while erecting this a bit in haste. Thanks for spotting.

amotl: Fixed per cd5c97c7f5d.

by-language/csharp-npgsql/DemoTypes.cs
kneth: As it is captured in https://github.com/crate/zk/issues/26, I think we are safe for now. Leaving a link to the issue will help us to remember.

kneth (approved):
LGTM

I would like to have links to existing issues for all `FIXME`s and `TODO`s so we don't forget what it is about.

PR #776, opened 2024-12-18
Npgsql: Use `Npgsql.CrateDB` instead of vanilla `Npgsql`

amotl:
## About
Probing better type support for Npgsql, or even trying to build it.

## Status
I don't know why `dotnet test` succeeds test runs in conditions I would expect it to fail. Anyway, on .NET 8, it just errors out like
```
    /path/to/cratedb-examples/by-language/csharp-npgsql/tests/DemoProgramTest.cs(25): error TESTERROR:
      demo.tests.DemoProgramTest.TestSystemQueryExample (1ms): Error Message: Class fixture type 'demo.tests.DatabaseFixture' threw in its constructor
      ---- Npgsql.NpgsqlException : Unknown message code: 0
```
`Npgsql.CrateDB` is outdated, as expected, and we need to investigate if it would yield a positive outcome if it would be resurrected.

### DIFF
--- a/by-language/csharp-npgsql/DemoProgram.cs
+++ b/by-language/csharp-npgsql/DemoProgram.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using CommandLine;
 using Npgsql;
+using Npgsql.CrateDb;
 using NpgsqlTypes;
 
 namespace demo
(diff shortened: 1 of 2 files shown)


amotl:
Closing this in favor of GH-777: Npgsql 8.x and 9.0 provide good support for mapping CrateDB's special data types to relevant types in .NET/Npgsql.

PR #777, opened 2024-12-19
Npgsql: Use `EnableDynamicJson` to unlock better container type mappings

amotl:
## About
This patch resolves a few FIXME admonitions introduced by a recent patch that started to validate all of CrateDB's data types with Npgsql.

- Yay: GH-772
- Not applicable: GH-776

## Details
Vanilla Npgsql provides good enough support to handle CrateDB's ARRAY and OBJECT types better than just plain strings. This patch demonstrates type mappings using three different variants:

- Native .NET `System.Collections.Generic.List` and `System.Collections.Generic.Dictionary` types.
- Native .NET `System.Text.Json.JsonDocument` type. [^1] Thanks, @simonprickett.
- Custom .NET POCO types.

## What's Inside
- Use `NpgsqlDbType.Json` explicitly where applicable, for example when communicating parameters using `AddWithValue()`.
- Use `EnableDynamicJson()` on the `NpgsqlDataSourceBuilder`, for POCO and `Dictionary` mappings that didn't work before.

## What's Next
Include relevant information into a dedicated documentation page, in order to educate readers properly in the spirit of "seeing is believing".

[^1]: Currently works with OBJECT for me, but not ARRAY.

### DIFF
--- a/by-language/csharp-npgsql/README.rst
+++ b/by-language/csharp-npgsql/README.rst
@@ -21,10 +21,8 @@ data provider for PostgreSQL, `crate-npgsql`_. CrateDB versions 4.2 and later
 work with the vanilla `Npgsql - .NET Access to PostgreSQL`_ driver without the
 need for a plugin.
 
-Please note that Npgsql 5 is not supported starting with CrateDB 4.8.4, you
-will need Npgsql 6 or newer.
-
-.NET 7, 8, and 9 are supported, .NET 3.1, 4.6, 5.0, and 6.0 may still work.
+The example program in this folder is validated on .NET 8 and 9,
+using Npgsql 8.x and 9.x.
 
 
 *****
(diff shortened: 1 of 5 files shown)


by-language/csharp-npgsql/README.rst
```diff
@@ -21,10 +21,8 @@ data provider for PostgreSQL, `crate-npgsql`_. CrateDB versions 4.2 and later
 work with the vanilla `Npgsql - .NET Access to PostgreSQL`_ driver without the
 need for a plugin.
 
-Please note that Npgsql 5 is not supported starting with CrateDB 4.8.4, you
-will need Npgsql 6 or newer.
-
-.NET 7, 8, and 9 are supported, .NET 3.1, 4.6, 5.0, and 6.0 may still work.
+The example program in this folder is validated on .NET 8 and 9,
```
kneth, line 24: Is it fair to say that the example will only work when using Npgsql 8.x and 9.x since `EnableDynamicJson()` was introduced on 8.0?

kneth (approved):

by-language/csharp-npgsql/README.rst
amotl: Yes. Do you think it needs an update?

kneth: Well, it is not incorrect what it says. And since Npgsl 7.x and earlier are not supported, it is unlikely that a user will try with anything by 8.x or 9.x

kneth: Let's leave it as it is

PR #782, opened 2024-12-21
Npgsql: Use `NpgsqlPoint` .NET type for marshalling `GEO_POINT` types. Explore communicating and marshalling GeoJSON types.

amotl:
## About
Explore CrateDB+Npgsql type support for geometry and geospatial types.

## Status
- `NpgsqlPoint` works well for `GEO_POINT` types. Thanks, @simonprickett.
- GeoJSON marshalling apparently needs to be done manually [^1].

[^1]: ... because `GEO_SHAPE` types are communicated as strings? I don't actually know the reason, but would like to investigate why it doesn't work fluently, optimally/optionally using the [PostGIS/GeoJSON Type Plugin](https://www.npgsql.org/doc/types/geojson.html)?

### DIFF
--- a/by-language/csharp-npgsql/DemoProgram.cs
+++ b/by-language/csharp-npgsql/DemoProgram.cs
@@ -17,12 +17,8 @@ namespace demo
                 {
                     var connString = $"Host={options.Host};Port={options.Port};SSL Mode={options.SslMode};" +
                                      $"Username={options.Username};Password={options.Password};Database={options.Database}";
-                    Console.WriteLine($"Connecting to {connString}\n");
 
-                    var dataSourceBuilder = new NpgsqlDataSourceBuilder(connString);
-                    dataSourceBuilder.EnableDynamicJson();
-                    await using var dataSource = dataSourceBuilder.Build();
-                    await using var conn = dataSource.OpenConnection();
+                    await using var conn = GetConnection(connString);
 
                     await DatabaseWorkloads.SystemQueryExample(conn);
                     await DatabaseWorkloads.BasicConversationExample(conn);
@@ -34,9 +30,32 @@ namespace demo
                     // await dwt.ArrayJsonDocumentExample();
                     await dwt.ObjectPocoExample();
                     await dwt.ArrayPocoExample();
+                    await dwt.GeoJsonTypesExample();
                     conn.Close();
                 });
 
+        }
+
+        public static NpgsqlConnection GetConnection(string connString)
+        {
+            Console.WriteLine($"Connecting to database: {connString}\n");
+
+            // Enable JSON POCO mapping and PostGIS/GeoJSON Type Plugin.
+            // https://www.npgsql.org/doc/types/json.html
+            // https://www.npgsql.org/doc/types/geojson.html
+            var dataSourceBuilder = new NpgsqlDataSourceBuilder(connString);
+
+            // Enable JSON POCO mapping Plugin.
+            // https://www.npgsql.org/doc/types/json.html
+            dataSourceBuilder.EnableDynamicJson();
+
+            // Enable PostGIS/GeoJSON Type Plugin.
+            // https://www.npgsql.org/doc/types/geojson.html
+            // dataSourceBuilder.UseGeoJson();
+
+            var dataSource = dataSourceBuilder.Build();
+            var conn = dataSource.OpenConnection();
+            return conn;
         }
 
         public class Options
--- a/by-language/csharp-npgsql/DemoTypes.cs
+++ b/by-language/csharp-npgsql/DemoTypes.cs
@@ -166,12 +166,27 @@ namespace demo
                 cmd.Parameters.AddWithValue("timestamp_tz", "1970-01-02T00:00:00+01:00");
                 cmd.Parameters.AddWithValue("timestamp_notz", "1970-01-02T00:00:00");
                 cmd.Parameters.AddWithValue("ip", "127.0.0.1");
+
+                // Container types
                 cmd.Parameters.AddWithValue("array", NpgsqlDbType.Json, new List<string>{"foo", "bar"});
                 cmd.Parameters.AddWithValue("object", NpgsqlDbType.Json, new Dictionary<string, string>{{"foo", "bar"}});
-                cmd.Parameters.AddWithValue("geopoint", new List<double>{85.43, 66.23});
-                // TODO: Check if `GEO_SHAPE` types can be represented by real .NET or Npgsql data types.
+
+                // Geospatial types
+
+                // GEO_POINT
+                // Alternatively to `NpgsqlPoint`, you can also use `List<double>{85.43, 66.23}`.
+                cmd.Parameters.AddWithValue("geopoint", new NpgsqlPoint(85.43, 66.23));
+
+                // GEO_SHAPE
+                // While `GEO_POINT` is transparently marshalled as `NpgsqlPoint`,
+                // `GEO_SHAPE` is communicated as scalar `string` type, using WKT or GeoJSON format.
+                // TODO: Possibly support transparently converging `GEO_SHAPE` to one of
+                //       `NpgsqlLSeg`, `NpgsqlBox`, `NpgsqlPath`, `NpgsqlPolygon`, `NpgsqlCircle`.
                 cmd.Parameters.AddWithValue("geoshape", "POLYGON ((5 5, 10 5, 10 10, 5 10, 5 5))");
+
+                // Vector type
                 cmd.Parameters.AddWithValue("float_vector", new List<double> {1.1, 2.2, 3.3});
+
                 cmd.ExecuteNonQuery();
             }
 
--- a/by-language/csharp-npgsql/DemoTypes.cs
+++ b/by-language/csharp-npgsql/DemoTypes.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Data;
 using System.Text.Json;
 using System.Threading.Tasks;
+using GeoJSON.Net.Geometry;
 using Newtonsoft.Json;
 using Npgsql;
 using NpgsqlTypes;
@@ -348,6 +349,102 @@ namespace demo
                 Console.WriteLine(obj[0]);
                 Console.WriteLine(obj[1]);
                 return obj;
+            }
+        }
+
+        public async Task InsertGeoJsonTyped()
+        {
+            /***
+             * Verify Npgsql PostGIS/GeoJSON Type Plugin with CrateDB.
+             * https://www.npgsql.org/doc/types/geojson.html
+             *
+             * TODO: Does not work yet, because CrateDB communicates GEO_SHAPE as string?
+             *       The error message is:
+             *
+             *       System.NotSupportedException : The NpgsqlDbType 'Geometry' isn't present in your
+             *       database. You may need to install an extension or upgrade to a newer version.
+             */
+            Console.WriteLine("Running InsertGeo");
+
+            // Insert single data point.
+            await using (var cmd = new NpgsqlCommand("""
+            INSERT INTO testdrive.example (
+                "geoshape"
+            ) VALUES (
+                @geoshape
+            );
+            """, conn))
+            {
+                var point = new Point(new Position(85.43, 66.23));
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Geometry, point);
+                cmd.ExecuteNonQuery();
+            }
+
+            // Flush data.
+            await RefreshTable();
+        }
+
+        public async Task InsertGeoJsonString()
+        {
+            /***
+             * Communicate GeoJSON types as strings, marshall from/to GeoJSON types manually.
+             */
+            Console.WriteLine("Running InsertGeoRaw");
+
+            // Insert single data point.
+            await using (var cmd = new NpgsqlCommand("""
+            INSERT INTO testdrive.example (
+                "geoshape"
+            ) VALUES (
+                @geoshape
+            );
+            """, conn))
+            {
+                var point = new Point(new Position(85.43, 66.23));
+                var poly = new Polygon([
+                    new LineString([
+                        new Position(longitude: 5.0, latitude: 5.0),
+                        new Position(longitude: 5.0, latitude: 10.0),
+                        new Position(longitude: 10.0, latitude: 10.0),
+                        new Position(longitude: 10.0, latitude: 5.0),
+                        new Position(longitude: 5.0, latitude: 5.0),
+                    ])
+                ]);
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Json, JsonConvert.SerializeObject(point));
+                cmd.ExecuteNonQuery();
+
+                cmd.Parameters.Clear();
+
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Json, JsonConvert.SerializeObject(poly));
+                cmd.ExecuteNonQuery();
+            }
+
+            // Flush data.
+            await RefreshTable();
+
+        }
+
+        public async Task<Point> GeoJsonTypesExample()
+        {
+            Console.WriteLine("Running GeoJsonTypesExample");
+
+            // Provision data.
+            await CreateTable();
+            // await InsertGeoJsonTyped();
+            await InsertGeoJsonString();
+
+            // Query back data.
+            await using (var cmd = new NpgsqlCommand("SELECT * FROM testdrive.example", conn))
+            await using (var reader = cmd.ExecuteReader())
+            {
+                reader.Read();
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                var obj = reader.GetFieldValue<JsonDocument>("geoshape");
+                var geoJsonObject = JsonConvert.DeserializeObject<Point>(obj.RootElement.ToString());
+                return (Point) geoJsonObject;
             }
         }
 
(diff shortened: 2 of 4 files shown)


amotl:
This patch offers a flakyness, where I don't know where it is coming from. It also happens on my machine. In a worst case, it could be CrateDB itself, but we would need to investigate to find out.

```
[xUnit.net 00:00:03.39]     demo.tests.DemoProgramTest.TestGeoJsonTypesExample [FAIL]
  Failed demo.tests.DemoProgramTest.TestGeoJsonTypesExample [173 ms]
  Error Message:
   Newtonsoft.Json.JsonReaderException : error parsing coordinates
---- Newtonsoft.Json.JsonReaderException : Unexpected character encountered while parsing value: [. Path 'coordinates', line 1, position 17.

   at demo.tests.DemoProgramTest.TestGeoJsonTypesExample() in /home/runner/work/cratedb-examples/cratedb-examples/by-language/csharp-npgsql/tests/DemoProgramTest.cs:line 222
```

https://github.com/crate/cratedb-examples/actions/runs/12444934016/job/34745706235?pr=782#step:7:105

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -336,6 +352,102 @@ public async Task<List<BasicPoco>> ArrayPocoExample()
...
+            """, conn))
+            {
+                var point = new Point(new Position(85.43, 66.23));
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Geometry, point);
+                cmd.ExecuteNonQuery();
+            }
+
+            // Flush data.
+            await RefreshTable();
+        }
```
amotl, line 385: It would be sweet to gain typed GeoJSON support, like the [Npgsql PostGIS/GeoJSON Type Plugin](https://www.npgsql.org/doc/types/geojson.html) might be providing it when talking to PostGIS. I don't know why it isn't working, the error message is:
```c#
System.NotSupportedException : The NpgsqlDbType 'Geometry' isn't present in your
database. You may need to install an extension or upgrade to a newer version.
```

Maybe it does not work, because CrateDB communicates `GEO_SHAPE` as exclusively as string when using the PostgrSQL wire protocol? Please advise if you see any options for improvements here.

/cc @seut, @surister, @kneth

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -336,6 +352,102 @@ public async Task<List<BasicPoco>> ArrayPocoExample()
...
+                ]);
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Json, JsonConvert.SerializeObject(point));
+                cmd.ExecuteNonQuery();
+
+                cmd.Parameters.Clear();
+
+                cmd.Parameters.AddWithValue("geoshape", NpgsqlDbType.Json, JsonConvert.SerializeObject(poly));
+                cmd.ExecuteNonQuery();
```
amotl, line 421: What works is to communicate GeoJSON data using the `NpgsqlDbType.Json` type, but it needs manual marshalling like `JsonConvert.SerializeObject(point)`.

Contrary to that, as mentioned above, the [Npgsql PostGIS/GeoJSON Type Plugin](https://www.npgsql.org/doc/types/geojson.html) enables to communicate .NET's GeoJSON types natively.

@simonprickett: Please let me know if you find any way do that already, which I might not have discovered yet. Thanks!

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -336,6 +352,102 @@ public async Task<List<BasicPoco>> ArrayPocoExample()
...
+            await using (var cmd = new NpgsqlCommand("SELECT * FROM testdrive.example", conn))
+            await using (var reader = cmd.ExecuteReader())
+            {
+                reader.Read();
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                var obj = reader.GetFieldValue<JsonDocument>("geoshape");
+                var geoJsonObject = JsonConvert.DeserializeObject<Point>(obj.RootElement.ToString());
+                return (Point?) geoJsonObject;
+            }
```
amotl, line 448: Dito: Manual procedures are currently needed when working with .NET's native GeoJSON types.

Here, the code uses `reader.GetFieldValue<JsonDocument>` for retrieval, and `JsonConvert.DeserializeObject<Point>(...)` for unmarshalling and type casting.

by-language/csharp-npgsql/tests/DemoProgramTest.cs
```diff
@@ -215,5 +212,21 @@ public async Task TestArrayPocoExample()
...
+            var conn = fixture.Db;
+
+            // Provision data.
+            var task = new DatabaseWorkloadsTypes(conn).GeoJsonTypesExample();
+            var point = await task.WaitAsync(TimeSpan.FromSeconds(0.5));
+
+            // Validate the outcome.
+            var coords = new Point(new Position(85.43, 66.23)).Coordinates;
+            Assert.Equal(coords.Latitude, point?.Coordinates.Latitude);
+            Assert.Equal(coords.Longitude, point?.Coordinates.Longitude);
```
amotl, line 227: Can't directly compare GeoJSON object instances. Need to run the comparison on the individual coordinates inside. 🤷

by-language/csharp-npgsql/DemoTypes.cs
kneth: Is PostGIS' `Geometry` equivalent to CrateDB's `geo_shape`? Would a simple alias in the server do the trick?

amotl: When talking about PostGIS, and looking at compatibility concerns, it is not just about types, but also, and mostly, about operations on them.

PostGIS unlocks [GDAL](https://en.wikipedia.org/wiki/GDAL), while CrateDB unlocks [JTS](https://en.wikipedia.org/wiki/JTS_Topology_Suite). Those are technically different animals, while they are still living in the same habitat. In this spirit, I figure that a simple alias will probably not be applicable, even if it would also be my dearest wish.

This doesn't mean we should not explore this area closer, how we could provide downstream compatibility, or at least a reasonable feature parity, possibly by other means.

amotl: > Would a simple alias in the server do the trick?

Maybe @seut has more insights into that. I will be so happy to also learn more about those details, and if they have been parts of any discussions in the past already.

seut: From reading the [PostGIS docs](https://postgis.net/docs/manual-2.4/using_postgis_dbmanagement.html#PostGIS_Geography), the generic `geometry` type standalone looks like it could be mapped to our `GEO_SHAPE` type as it serves as a generic type for all concrete spatial types the same like our `geo_shape` does. But as far as I read, the PostGIS also allows to specify a concrete geometry subset on table definition, e.g. `geometry(LINESTRING)`.  This isn't possible at CrateDB, we cannot limit the allowed concrete shape of a geometry value. 
So we maybe could alias the PostGIS `geometry` type without a concrete spatial type definition, but even so we'd need to do some (extensive) testing to ensure that this works as expected (in terms of SQL and PG compatibility), especially, the query/filter behaviour.
I suggest to open a feature request at CrateDB to implement the `geometry` data type with the alias of `geo_shape` as a possible solution.

amotl: @seut: Thanks for your swift elaborations, and Happy New Year.
@kneth: Can I humbly ask you to follow @seut's advise and carry your proposal forward into a corresponding feature request ticket at crate/crate?

kneth: @amotl @seut I have created https://github.com/crate/crate/issues/17187

amotl: Thank you. 👍

amotl:
Do you have any objections to merge this patch, @simonprickett and @kneth?

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -336,6 +352,102 @@ public async Task<List<BasicPoco>> ArrayPocoExample()
...
+            // Query back data.
+            await using (var cmd = new NpgsqlCommand("SELECT * FROM testdrive.example", conn))
+            await using (var reader = cmd.ExecuteReader())
+            {
+                reader.Read();
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                var obj = reader.GetFieldValue<JsonDocument>("geoshape");
+                var geoJsonObject = JsonConvert.DeserializeObject<Point>(obj.RootElement.ToString());
+                return (Point?) geoJsonObject;
```
seut, line 447: Reading data is not directly related to how data is ingested. Shouldn't the marshalling work when enabling GeoJSON by `dataSourceBuilder.UseGeoJson();`? Afaik the output JSON should be a valid GeoJSON format, at least for simple structures like Points.

by-language/csharp-npgsql/DemoProgram.cs
```diff
@@ -34,11 +30,34 @@ await Parser.Default.ParseArguments<Options>(args)
...
+            // https://www.npgsql.org/doc/types/geojson.html
+            var dataSourceBuilder = new NpgsqlDataSourceBuilder(connString);
+
+            // Enable JSON POCO mapping Plugin.
+            // https://www.npgsql.org/doc/types/json.html
+            dataSourceBuilder.EnableDynamicJson();
+
+            // Enable PostGIS/GeoJSON Type Plugin.
+            // https://www.npgsql.org/doc/types/geojson.html
+            // dataSourceBuilder.UseGeoJson();
```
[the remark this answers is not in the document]
amotl: Thanks for your suggestions, here and below. I will investigate them and report back.

kneth:
I removed requested reviewers for now, and you can add new - or same - reviewers when the PR moved out of draft.

by-language/csharp-npgsql/DemoTypes.cs
amotl: It looks like CrateDB sometimes returns a `Point`, which satisfies the current code, aka. it succeeds in this case.
```json
{"coordinates":[66.23,85.43],"type":"Point"}
```

In the other case when this fails, CrateDB returned that:
```json
{"coordinates":[[[5.0,5.0],[5.0,10.0],[10.0,10.0],[10.0,5.0],[5.0,5.0]]],"type":"Polygon"}
```

by-language/csharp-npgsql/DemoTypes.cs
amotl: Unfortunately, the marshalling does not work, see conversation [here](https://github.com/crate/cratedb-examples/pull/782#pullrequestreview-2518693533).

by-language/csharp-npgsql/DemoProgram.cs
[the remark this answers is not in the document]
amotl: I've tried to amend the corresponding [INSERT statement](https://github.com/crate/cratedb-examples/pull/782#pullrequestreview-2518693533), but found that it isn't even executed. The program already fails before, probably based on a type introspection query included in those bunch of statements I've traced using `ctk tail`.

<details>
<summary>Npgsql introspection SQL</summary>

```sql
SELECT version();


SELECT ns.nspname,
       t.oid,
       t.typname,
       t.typtype,
       t.typnotnull,
       t.elemtypoid
FROM
  (-- Arrays have typtype=b - this subquery identifies them by their typreceive and converts their typtype to a
 -- We first do this for the type (innerest-most subquery), and then for its element type
 -- This also returns the array element, range subtype and domain base type as elemtypoid
 SELECT typ.oid,
        typ.typnamespace,
        typ.typname,
        typ.typtype,
        typ.typrelid,
        typ.typnotnull,
        typ.relkind,
        elemtyp.oid AS elemtypoid,
        elemtyp.typname AS elemtypname,
        elemcls.relkind AS elemrelkind,
        CASE
            WHEN elemproc.proname='array_recv' THEN 'a'
            ELSE elemtyp.typtype
        END AS elemtyptype ,
        typ.typcategory
   FROM
     (SELECT typ.oid,
             typnamespace,
             typname,
             typrelid,
             typnotnull,
             relkind,
             typelem AS elemoid,
             CASE
                 WHEN proc.proname='array_recv' THEN 'a'
                 ELSE typ.typtype
             END AS typtype,
             CASE
                 WHEN proc.proname='array_recv' THEN typ.typelem
                 WHEN typ.typtype='r' THEN rngsubtype
                 WHEN typ.typtype='m' THEN
                        (SELECT rngtypid
                         FROM pg_range
                         WHERE rngmultitypid = typ.oid)
                 WHEN typ.typtype='d' THEN typ.typbasetype
             END AS elemtypoid ,
             typ.typcategory
      FROM pg_type AS typ
      LEFT JOIN pg_class AS cls ON (cls.oid = typ.typrelid)
      LEFT JOIN pg_proc AS proc ON proc.oid = typ.typreceive
      LEFT JOIN pg_range ON (pg_range.rngtypid = typ.oid)) AS typ
   LEFT JOIN pg_type AS elemtyp ON elemtyp.oid = elemtypoid
   LEFT JOIN pg_class AS elemcls ON (elemcls.oid = elemtyp.typrelid)
   LEFT JOIN pg_proc AS elemproc ON elemproc.oid = elemtyp.typreceive) AS t
JOIN pg_namespace AS ns ON (ns.oid = typnamespace)
WHERE (typtype IN ('b',
                   'r',
                   'm',
                   'e',
                   'd')
       OR -- Base, range, multirange, enum, domain
 (typtype = 'c'
  AND relkind='c')
       OR -- User-defined free-standing composites (not table composites) by default
 (typtype = 'p'
  AND typname IN ('record',
                  'void',
                  'unknown'))
       OR -- Some special supported pseudo-types
 (typtype = 'a'
  AND (-- Array of...
 elemtyptype IN ('b',
                 'r',
                 'm',
                 'e',
                 'd')
       OR -- Array of base, range, multirange, enum, domain
 (elemtyptype = 'p'
  AND elemtypname IN ('record',
                      'void'))
       OR -- Arrays of special supported pseudo-types
 (elemtyptype = 'c'
  AND elemrelkind='c')-- Array of user-defined free-standing composites (not table composites) by default
)))
ORDER BY CASE
             WHEN typtype IN ('b',
                              'e',
                              'p') THEN 0 -- First base types, enums, pseudo-types

             WHEN typtype = 'c' THEN 1 -- Composites after (fields loaded later in 2nd pass)

             WHEN typtype = 'r' THEN 2 -- Ranges after

             WHEN typtype = 'm' THEN 3 -- Multiranges after

             WHEN typtype = 'd'
                  AND elemtyptype <> 'a' THEN 4 -- Domains over non-arrays after

             WHEN typtype = 'a' THEN 5 -- Arrays after

             WHEN typtype = 'd'
                  AND elemtyptype = 'a' THEN 6 -- Domains over arrays last

         END;

-- Load field definitions for (free-standing) composite types

SELECT typ.oid,
       att.attname,
       att.atttypid
FROM pg_type AS typ
JOIN pg_namespace AS ns ON (ns.oid = typ.typnamespace)
JOIN pg_class AS cls ON (cls.oid = typ.typrelid)
JOIN pg_attribute AS att ON (att.attrelid = typ.typrelid)
WHERE (typ.typtype = 'c'
       AND cls.relkind='c')
  AND attnum > 0
  AND -- Don't load system attributes
 NOT attisdropped
ORDER BY typ.oid,
         att.attnum;

-- Load enum fields

SELECT pg_type.oid,
       enumlabel
FROM pg_enum
JOIN pg_type ON pg_type.oid=enumtypid
ORDER BY oid,
         enumsortorder;
```

</details>

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -166,12 +167,27 @@ INSERT INTO testdrive.example (
...
+
+                // Geospatial types
+
+                // GEO_POINT
+                // Alternatively to `NpgsqlPoint`, you can also use `List<double>{85.43, 66.23}`.
+                cmd.Parameters.AddWithValue("geopoint", new NpgsqlPoint(85.43, 66.23));
+
+                // GEO_SHAPE
+                // While `GEO_POINT` is transparently marshalled as `NpgsqlPoint`,
+                // `GEO_SHAPE` is communicated as scalar `string` type, using WKT or GeoJSON format.
```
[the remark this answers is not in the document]
amotl: Sorry, it didn't work.

by-language/csharp-npgsql/DemoTypes.cs
```diff
@@ -336,6 +353,108 @@ public async Task<List<BasicPoco>> ArrayPocoExample()
...
+            await using (var reader = cmd.ExecuteReader())
+            {
+                reader.Read();
+                // TODO: Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
+                //       Currently, `InsertGeoJsonTyped` does not work yet.
+                var obj = reader.GetFieldValue<JsonDocument>("geoshape");
+                var payload = obj.RootElement.ToString();
+                var geoJsonObject = JsonConvert.DeserializeObject<Point>(payload);
+                return (Point?) geoJsonObject;
+            }
```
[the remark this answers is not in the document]
amotl: That would cause other warnings or more elaborate downstream error handling. We don't want to introduce such details here, because the focus is on concise demo examples.

kneth (approved):
Let's merge it but create an issue for the todos like

> Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
> Currently, `InsertGeoJsonTyped` does not work yet.

by-language/csharp-npgsql/DemoTypes.cs
amotl: @kneth suggested to track this into the future by using a dedicated ticket. Thanks.

> Can GEO_SHAPE types be directly marshalled to a .NET GeoJSON type?
> Currently, `InsertGeoJsonTyped` does not work yet.

@coderabbitai: Please pick up some context from relevant conversations here, and converge that into a dedicated issue for us. Thanks.

amotl:
We've created a ticket to track certain type support into the future, thanks.
- GH-1299

PR #791, opened 2025-01-09
Updated the wind farm Jupyter notebook for new data links.

simonprickett:
Changed the wind farm Jupyter notebook to use new data links as it is no longer wise to directly access LFS files from GitHub.


surister (approved):
I got 
image
while checking the links in firefox, not sure if it actually affects the notebook, if it doesn't then LGTM

simonprickett:
> I got image while checking the links in firefox, not sure if it actually affects the notebook, if it doesn't then LGTM

I think this is because it's really JSONL so the broswer won't render it.

PR #821, opened 2025-01-31
Example showing prepared statements with pg8000

hlcianfagna:

Adds a new example showing how to use prepared statements with CrateDB


amotl (approved):
Thanks!

NB: You can now use Markdown everywhere.

PR #855, opened 2025-03-21
Rust: Add minimal example

amotl:
## About

Add minimal Rust example from [crate-qa](https://github.com/crate/crate-qa/tree/master/tests/client_tests/rust) to make a start, adding a miniature rig for further explorations.

/cc @kneth

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -83,6 +83,11 @@ updates:
     schedule:
       interval: "daily"
 
+  - directory: "/by-language/rust"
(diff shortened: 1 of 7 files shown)


surister (approved):
The only thing to comment here is that by unwrapping every statement we are letting panics run free, good practice would be to handle errors explicitly, but its just a minimal example, I'm fine with it.

amotl:
Thank you very much. Please use this miniature rig to improve it as we go. Any contribution is welcome, also about improving error handling, even if it's just "little".

PR #860, opened 2025-03-22
Testcontainers for Java: Update software versions

amotl:
Use CrateDB 5.10 and PostgreSQL 17.

_Not **every** thing is automatic. /cc @kneth_


surister (approved):

kneth:
Could it be a constant so we only have to edit one place?

PR #861, opened 2025-03-22
Java: Also test against most recent Java 24

amotl:
Add Java 24 to the test matrix of the Java snippets.


amotl:
Gradle does not provide support Java 24 yet, see [Compatibility Matrix](https://docs.gradle.org/current/userguide/compatibility.html).

amotl:
> Gradle does not provide support for Java 24 yet.

It looks like it works now.

matriv (approved):

PR #863, opened 2025-03-24
Improve dbt materialization example

hlcianfagna:

This works around https://github.com/crate/dbt-cratedb2/issues/10#issuecomment-2747796826


amotl (approved):
Thanks!

PR #879, opened 2025-04-01
ConnectorX: Hello, World!

amotl:
### About

> [ConnectorX](https://sfu-db.github.io/connector-x/) is a Python module to access relational databases using Rust and building upon Apache Arrow. It aims to be efficient, and focuses on the read path only.

### README

Display the rendered version at [Using CrateDB with ConnectorX](https://github.com/crate/cratedb-examples/blob/connectorx/by-language/python-connectorx/README.md).

### Details

The patch just contains a basic example within [`demo.py`](https://github.com/crate/cratedb-examples/blob/connectorx/by-language/python-connectorx/demo.py), which needs to be expanded.

/cc @karynzv, @hlcianfagna, @hammerhead, @wierdvanderhaar, @WalBeh


amotl:
Merged per "ack-by-timeout". ;]

PR #898, opened 2025-04-16
CrateDB SSL: Add client programs using `sslmode=require` and enable testing on CI

amotl:
## About

Provide example client programs using the canonical `?sslmode=require` query parameter, which means to connect to SSL-based servers, but to turn off certificate validation.

## References

- https://github.com/crate/sqlalchemy-cratedb/pull/202
- https://github.com/crate/cratedb-toolkit/pull/408
- https://github.com/crate/crate/pull/5719

/cc @WalBeh


hlcianfagna (commented):
Looks good as an improvement on the [original PR](https://github.com/crate/cratedb-examples/pull/825) which I had not seen. I have a comment however which maybe you could address in the README? unless I am missing something this example is not generating fresh self signed certificates but using the ones uploaded in the repo so anyone spinning an environment from this example will be using the same encryption keys.

amotl:
Thanks. I've added a new prominent section to the README:

> ## Caveat
>
> Note this example is not generating fresh self-signed certificates,
but uses the ones uploaded to the repository. In this spirit, anyone
spinning an environment from this example will be using the same
encryption keys. A future iteration might improve this, possibly by
using [minica].

[minica]: https://github.com/jsha/minica

hlcianfagna (approved):

PR #956, opened 2025-05-29
Fix CrateDB Cloud URI

hlcianfagna:
## Checklist


amotl (approved):
Thanks for spotting!

hlcianfagna:
> question: ssl=true works even in local without any certificates set up?

No, it returns `SSL: WRONG_VERSION_NUMBER`

PR #990, opened 2025-06-17
Apache Superset: Improve probing including pre-release versions

amotl:
## About
[Apache Superset 4.1.3rc1](https://pypi.org/project/apache-superset/4.1.3rc1/) released on May 6 could make our CI happy again.

## References
- https://github.com/apache/superset/issues/33162#issuecomment-2981902735

### DIFF
--- a/.github/workflows/application-apache-superset.yml
+++ b/.github/workflows/application-apache-superset.yml
@@ -40,9 +40,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-22.04 ]
-        superset-version: [ "3.*", "4.*" ]
+        superset-version: [ "==3.*", "==4.*", "==5.*" ]
         python-version: [ "3.9", "3.11" ]
         cratedb-version: [ 'nightly' ]
+        exclude:
+          - superset-version: "==5.*"
+            python-version: "3.9"
 
     services:
       cratedb:
@@ -75,7 +78,7 @@ jobs:
 
       - name: Install Apache Superset ${{ matrix.superset-version }}
         run: |
-          pip install 'apache-superset==${{ matrix.superset-version }}'
+          pip install --pre 'apache-superset${{ matrix.superset-version }}'
 
       - name: Validate application/apache-superset
         run: |


mistercrunch:
Trying to think of the best way to either a) offer some LTS where it matters, or b) marking older versions as "not supported, may have security issues". Realistically given the focus of current maintainers, support previous/current/next major is already challenging.

The main issue is around supporting a large tree of dependency both on the backend and frontend. Some tooling offers issue reporting on the dep tree, flagging individual librairies down the tree that have issues, but the whole thing is challenging to manage. Realistically the only way to stay on top of it is automation/tooling, using things like Snyk and @dependabot / @supersetbot to help in the process. Now given the largely manual process dictated by the ASF the includes humans voting release through email, things get pretty hard.

Release automation / tooling would really help. It's probably a matter of reinvesting in custom tooling like https://github.com/apache-superset/cherrytree and https://github.com/apache-superset/supersetbot - or go with the latest AI-powered tooling to try and support/automate release management. Until AI takes over, this would require a significant amount of good old human-powered engineering.

amotl:
> Now given the largely manual process dictated by the ASF the includes humans voting release through email, things get pretty hard.

Oh wow. I would never have imagined such obstacles. Thanks for your extensive elaboration.

> Another thought on the current release process. It seems possible for maintainers to allow people like you who want older releases to be pushed forward, to open PRs against release branches, and allow release branches to be ahead of official release (that require email voting, issue of multiple release candidates, and finally release publishing).
>
> So maybe we allow for release branches to evolve, so say you target the 3.1 branch with the marshmallow boundary PR, we merge it, and they 3.1 is ahead of the latest official release, say `3.1.3`, and maybe eventually - ideally through some automation, `3.1.4rc1` gets submitted for release. Or maybe 3.1 just keeps pilling up LTS-type fixes and remains supported longer term.

Yes indeed. On the [CrateDB repository](https://github.com/crate/crate), we are maintaining the 4.x and 5.x branches, dropped support for 3.x, and are currently transitioning to 6.x. All active branches of major versions accept changes, and we are employing Mergify to support fluent backports of PRs without needing to do that manually.

Of course, it is naturally up to you which major versions you will be supporting. Our thinking just was that because 5.x is still in its infancy, we felt it would be obvious to keep 3.x and 4.x active, but we haven't been able to find a 3.x branch to direct our dependency fix to.

> Release automation / tooling [around dependency management] would really help.

I don't disagree, but we are really looking at "just accepting patches" on particular spots where things go south, in this case with Marshmallow 4.x. This does not necessarily mean to redesign the whole dependency management machinery. I absolutely hear you on maintainer burnout topics, related to all of that. I think no matter what, KISS approaches are the only ones that are bearable. 🍀

mistercrunch:
If we wanted to work around the hurdles of "official ASF releases" which require an email vote, tarballs on a ASF-hosted SVN and all, people can/could point their build system to our release branches. In our case, the minor (ie 4.1) is a branch, and the official release (ie 4.1.1) is a tag on that branch. In theory we could move faster if you build on 4.1 and we can easily target PRs to those branches (say fixing the `marshmallow<=4.x` ceiling thing). Then you're still on 4.1 but ahead of the latest ASF release. It becomes a bit of a moving target but might achieve your goals.

amotl:
Superset `master` has been fixed.
- https://github.com/apache/superset/pull/33978

Now the patch is only missing for the `5.0` branch.
- https://github.com/apache/superset/issues/33162#issuecomment-3092390913

amotl:
- Testing against pre-release versions makes sense, but I can't adjust this branch any longer.
- @bgunebakan applied a few fixes the other day: GH-1891. Thanks!

amotl:
@bgunebakan or @kneth: Please delete my branch `superset-4.1.3rc1`. 🙇

PR #1185, opened 2025-11-02
Gradle 9: Fix build for by-language/java-jooq

amotl:
## About

Mitigate some problems to fix the build after Gradle 9 was released.

## Problems

```
A problem occurred evaluating script.
> org/gradle/api/plugins/JavaPluginConvention
```

```
Could not get unknown property 'destinationDir' for task
':compileJava' of type org.gradle.api.tasks.compile.JavaCompile.
```

```
class file has wrong version 65.0, should be 61.0
```

```
Could not resolve nu.studer:gradle-jooq-plugin:10.1.1
Dependency requires at least JVM runtime version 21.
This build uses a Java 17 JVM.
```

### DIFF
--- a/.github/workflows/lang-java-jooq.yml
+++ b/.github/workflows/lang-java-jooq.yml
@@ -39,7 +39,7 @@ jobs:
         os: [ 'ubuntu-latest' ]
         java-version: [
           '21',
-          '24',
+          '25',
         ]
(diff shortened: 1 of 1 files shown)


matriv (approved):
Maybe already add java 25 (LTS)?

PR #1188, opened 2025-11-02
Python/turbodbc: Use package from Git repository

amotl:
## About
An attempt to harmonize recent pyarrow with turbodbc.

## References
- GH-1177


amotl:
- See: https://github.com/crate/cratedb-examples/pull/1177#issuecomment-3477950186

PR #1203, opened 2025-11-04
Gradle: Remove wrappers

amotl:
Gradle wrappers include a bunch of jar files and shell programs. Apparently, invoking `gradle test` also works well without them. Why have they been added before?


matriv:
I don't agree with this change. The gradle wrapper is the way to have a consistent run, independently of the gradle vesion installed on CI machines. If the gradle is upgraded there we might end up with maintenance cycles, every time this happens.

amotl:
I see. I was wondering why the update to Gradle 9 tripped so many builds, so I was questioning if the mechanism is effective. Maybe we have been using it wrong?

amotl:
>> so I was questioning if the mechanism is effective
>
> Maybe we have been using it wrong?

I see, the [relevant code](https://github.com/pyveci/pueblo/blob/v0.0.11/pueblo/ngr/runner.py#L257-L266) has a comment that says:

```python
# When `gradle` is installed, wipe the
# existing wrapper, and generate a new one.
```

In this spirit, the test runner currently encourages users to modernize ;], which isn't actually a bad thing. However, the runner should probably follow a different default code path, **actually using the wrapper** mostly to validate it, and **only optionally** skip it by generating a new one using `gradle wrapper`.

- https://github.com/pyveci/pueblo/issues/183

PR #1342, opened 2026-01-10
ingestr: Add Elasticsearch demo

amotl:
It's an ingestr demo that imports data from Elasticsearch into CrateDB. It doesn't do much about verifying data, it is merely about validating the workflow that it runs to completion successfully.

### DIFF
--- a/application/ingestr/elasticsearch-compose.yml
+++ b/application/ingestr/elasticsearch-compose.yml
@@ -96,3 +96,18 @@ services:
       crash --hosts cratedb -c "SELECT * FROM doc.example LIMIT 10"
     deploy:
       replicas: 0
+
+  # Verify data in CrateDB.
+  cratedb-verify:
+    build:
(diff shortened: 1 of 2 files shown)


application/ingestr/elasticsearch-compose.yml
```diff
@@ -0,0 +1,98 @@
...
+  # Create Elasticsearch index.
+  es-create-index:
+    image: docker.io/alpine/httpie
+    command: PUT http://elasticsearch:9200/example
+    deploy:
+      replicas: 0
+
+  # Import data from CSV into Elasticsearch.
+  es-import:
+    image: ghcr.io/bruin-data/ingestr:v0.14.116
```
amotl: Let's bump to v0.14.117 before merging this patch when it's released.

- https://github.com/bruin-data/ingestr/issues/487#issuecomment-3739170790

PR #1494, opened 2026-03-24
pgx: Start validating PostgreSQL protocol version 3.2

amotl:
## About
PostgreSQL 18 added a new protocol version, version 3.2. pgx v5.9.0 added support for it, optionally to be enabled/configured using the new `min_protocol_version` and `max_protocol_version` options. By default, pgx will still use protocol version 3.0 like before.

## Details
The corresponding announcement in the pgx v5.9.0 change log sounds interesting, it [says](https://github.com/jackc/pgx/blob/v5.9.0/CHANGELOG.md?plain=1#L6-L7):

> It significantly reduces the amount of network traffic when using prepared statements (which are used automatically by default) by avoiding unnecessary Describe Portal messages. This also reduces local memory usage.

## References
- https://github.com/crate/crate/issues/19175
- https://github.com/crate/cratedb-examples/pull/1507
- Upstream feature: https://github.com/jackc/pgx/issues/2489
- Currently blocked by GH-1490, see https://github.com/crate/crate-clients-tools/issues/306.

### DIFF
--- a/.github/workflows/lang-go-pgx.yml
+++ b/.github/workflows/lang-go-pgx.yml
@@ -29,14 +29,18 @@ jobs:
   test:
     name: "
      Go: ${{ matrix.go-version }}
+     Protocol: ${{ matrix.postgresql-protocol-version }}
      CrateDB: ${{ matrix.cratedb-version }}
-     on ${{ matrix.os }}"
+    "
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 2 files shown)


amotl:
CrateDB doesn't talk protocol version 3.2 yet. Currently, this isn't a problem at all, because protocol negotiation is in place. The optimization is only relevant for pgx, not in general.

PR #1507, opened 2026-03-24
jdbc: Start validating PostgreSQL protocol version 3.2

amotl:
## About
PostgreSQL 18 added a new protocol version, version 3.2. The PostgreSQL JDBC driver was accompanied with the `protocolVersion` option that can be used within the database connection URL. By default, PostgreSQL JDBC will still use protocol version 3.0 like before.

## Details
The corresponding announcement in the pgx v5.9.0 change log sounds interesting, it [says](https://github.com/jackc/pgx/blob/v5.9.0/CHANGELOG.md?plain=1#L6-L7):

> It significantly reduces the amount of network traffic when using prepared statements (which are used automatically by default) by avoiding unnecessary Describe Portal messages. This also reduces local memory usage.

## References

- https://github.com/crate/crate/issues/19175
- https://github.com/crate/cratedb-examples/pull/1494

### DIFF
--- a/.github/workflows/lang-java-maven.yml
+++ b/.github/workflows/lang-java-maven.yml
@@ -32,6 +32,7 @@ jobs:
   test:
     name: "
      Java: ${{ matrix.java-version }}
+     Protocol: ${{ matrix.postgresql-protocol-version }}
      CrateDB: ${{ matrix.cratedb-version }}
      on ${{ matrix.os }}"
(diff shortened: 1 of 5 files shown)


amotl:
CrateDB doesn't talk protocol version 3.2 yet. Currently, this isn't a problem at all, because protocol negotiation is in place. The optimization is only relevant for pgx, not in general.

Issue #1779, opened 2026-06-11
A less aggressive dependabot


kneth:
Currently, dependabot is creating PRs faster than we can review and merge. To make dependabot less agresive, we need to consider the following to settings

* schedule
* versioning-strategy

amotl:
Hi. Please don't adjust the schedule. Having daily runs gives us the chance to react early to relevant changes, not only after a month when more users might be impacted from something that went south.

Tuning the test frequency down would defy this very important purpose of the repository. If you don't like any version bumping policy currently applied, please adjust that, or migrate away from Dependabot.

Dependabot was starting to go crazy about bumping lower bound version numbers only recently. Maybe this is what you are referring to as "aggressive"?

- https://github.com/dependabot/dependabot-core/issues/14798#issuecomment-4450926177

Upstream people are working on improving the situation:

- https://github.com/dependabot/dependabot-core/pull/14825

Otherwise, we can also switch from Dependabot to Renovate, which successfully accepts the "widen" version bumping strategy.

amotl:
> Dependabot was starting to go crazy about bumping lower bound version numbers only recently. 

In this spirit, please also close all those PRs. They are wrong.

<img width="641" height="370" alt="image" src="https://github.com/user-attachments/assets/608ecd04-e234-4eec-b414-4b260232b118" />

amotl:
> Otherwise, we can also switch from Dependabot to Renovate, which successfully accepts the "widen" version bumping strategy.

This will lead to less Dependabot traffic as long as Dependabot does not re-adopt the "widen" policy. For a blueprint, see CTK's [renovate.json5](https://github.com/crate/cratedb-toolkit/blob/main/renovate.json5) configuration file.

amotl:
> Currently, dependabot is creating PRs faster than we can review and merge.

Holding up to a diligent daily routine to process all signals is crucial. Otherwise, all the benefits like catching regressions in CrateDB and reacting early to relevant hiccups will become moot. If you start training, possible drag will quickly go away and the benefits will shine through. You can also use the venerable [build status page](https://cratedb.com/docs/guide/integrate/status.html) as a tool for this purpose.

amotl:
Hi. This spot is another good example why scheduling nightly runs and Dependabot on a daily interval is advisable to catch any kinds of regressions through package updates or other moving targets.

- crate/sqlalchemy-cratedb#284

amotl:
Hi again,

[...]

PR #1781, opened 2026-06-11
A less agressive dependabot

kneth:
## Summary of the changes / Why this is an improvement

We change the frequency from daily to monthly, and the [versioning strategy](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#versioning-strategy--) to `auto`.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): Fixes #1779

PR #1802, opened 2026-06-16
Dependabot: Remove `widen` strategy again, it is rejected now

amotl:
## Problem

We found this message on a recent CI run.

> The property '#/updates/1/versioning-strategy' value "widen" did not match one of the following values: lockfile-only, auto, increase, increase-if-necessary.

-- https://github.com/crate/cratedb-examples/runs/81021738982

## References
- Tangentially related: GH-1779

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,7 +18,6 @@ updates:
 
   - directory: "/"
     package-ecosystem: "pip"
-    versioning-strategy: "widen"
     schedule:
       interval: "daily"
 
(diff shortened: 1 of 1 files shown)


amotl:
Please review and merge if you agree. /cc @bgunebakan, @florinutz

bgunebakan:
There is already an open PR for changing `widen` to `auto` So we don't need to delete that in separate PR, It will be solved in #1781

amotl:
Thanks. However, that other PR also mixes in scheduling changes that might be detrimental. Please select any of both patches at your disposal. Personally, I think it is better to remove the setting completely again so it does not create any future confusion. Apologies for introducing it the other day.

amotl:
Hi again. This patch will not improve anything for the better. Let's close it and refer to another one instead that really makes a difference in this regard.

- GH-1841

PR #1839, opened 2026-06-19
Python/SQLAlchemy: Stop testing experimental features

amotl:
## About
To increase the stability of integration tests touched hereby, leave testing against the stale [`amo/postgresql-async`](https://github.com/crate/sqlalchemy-cratedb/pull/11) development branch behind.

## Details
`amo/postgresql-async` is a development branch, and it can no longer easily be refreshed. It is better to demonstrate only GA features anyway, and flag the experimental ones appropriately.

## Reason

All those CI jobs are failing otherwise.

- crate/cratedb-examples#1799
- crate/cratedb-examples#1808
- crate/cratedb-examples#1816

PR #1867, opened 2026-06-24
Switch sqlalchemy-cratedb from dev branch to released 0.43.1

bgunebakan:
The **SQLAlchemy** and **pandas** nightly CI jobs were failing with:

```python
    ParsingException[line 1:71: no viable alternative at input
      'SELECT table_name FROM information_schema.tables WHERE table_schema = %']
```

Both examples were pinned to the `amo/postgresql-async` dev branch of `sqlalchemy-cratedb`, which adds a `_format_query` helper that rewrites `?` placeholders to `%s` when the dialect's `paramstyle` is `pyformat`. This branch is old and does not have features that used in current package.

This update will fix the pipeline and will help to test with latest packages.

## Checklist

### DIFF
--- a/by-language/python-sqlalchemy/test.py
+++ b/by-language/python-sqlalchemy/test.py
@@ -8,7 +8,8 @@ def run(command: str):
 
 
 def test_insert_efficient_multirow():
-    insert_records = 25_000
+    # CrateDB enforces statement_max_length=262144; keep well under that limit.
+    insert_records = 5_000
     cmd = f"time python insert_efficient.py cratedb multirow {insert_records}"
     run(cmd)
 
--- a/by-language/python-sqlalchemy/test.py
+++ b/by-language/python-sqlalchemy/test.py
@@ -26,17 +26,20 @@ def test_insert_efficient_unknown(capfd):
     out, err = capfd.readouterr()
     assert "ValueError: Unknown variant: unknown" in err
 
-
+# TODO: Remove skip when psycopg/asyncpg dialect is released in sqlalchemy-cratedb
+@pytest.mark.skip(reason="psycopg/asyncpg dialect not yet in released sqlalchemy-cratedb")
 def test_sync_table():
     cmd = "time python sync_table.py urllib3 psycopg"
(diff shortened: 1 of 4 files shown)


amotl:
Hi Bilal. Thank you. This patch roughly looks like a duplicate of this other one?

- GH-1839

by-language/python-sqlalchemy/test.py
```diff
@@ -8,13 +8,14 @@ def run(command: str):
...
-    insert_records = 25_000
+    # CrateDB enforces statement_max_length=262144; keep well under that limit.
+    insert_records = 5_000
     cmd = f"time python insert_efficient.py cratedb multirow {insert_records}"
     run(cmd)
 
 
 def test_insert_efficient_batched():
-    insert_records = 50_000
+    insert_records = 5_000
```
amotl, line 18: Thank you for adding this. I've seen the same thing, and it might be a huge red flag against the recent change in crate-python, so we might not want to have this slip unnoticed? In GH-1839, I am deliberately using `crate>=2.1.2,<2.2` and `sqlalchemy-cratedb[all]<0.43`, in order to surface this discovery on behalf of a later PR and flag it for review more prominently.

I think the change is significant for some users who use the manyrows marshalling style (default with pandas), as it will decrease available bulk sizes in standard communication paths with CrateDB by a huge factor.

In this spirit, I was planning two phases here:
a) Restore compatibility (phase out experimental features) with GH-1839.
b) Surface the bulk size regression with a later patch that upgrades package versions.
c) Present the regression to other colleagues.

wdyt?

kneth (approved):

by-language/python-sqlalchemy/test.py
amotl: We are tracking this observation here.

- https://github.com/crate/crate-python/issues/818

bgunebakan: I explained details about this on the related issue page, This is not a bug so there is nothing to do. Just adjusting example codes are enough. Thanks.

